PR #16, opened 2020-09-08
Removed duplicate param comment.

cousinitt:
For the constructor of MagneticSensorAnalog (in the comments) there was a duplicated param, there were also two other arguments to the constructor that were not described (If I figure them out, I'll put them in?)


askuric:
Well spotted. :D

PR #17, opened 2020-09-08
Added c++11 "override" labels to overriden Sensor methods.

cousinitt:
The override keyword causes the compiler to bark at you if when you specify a method override with a spelling error.
(As of 2014-ish, the Arduino IDE was supposed to support C++11 ... I haven't actually tried this out yet in my Arduino IDE ... I'm on a different computer).
No feelings will be hurt if you reject this one.

### DIFF
--- a/src/Encoder.h
+++ b/src/Encoder.h
@@ -59,31 +59,31 @@ class Encoder: public Sensor{
 
     // Abstract functions of the Sensor class implementation
     /** get current angle (rad) */
-    float getAngle();
+    float getAngle() override;
     /**  get current angular velocity (rad/s) */
-    float getVelocity();
+    float getVelocity() override;
     /** 
      *  set current angle as zero angle 
      * return the angle [rad] difference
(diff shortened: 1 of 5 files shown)


askuric:
I am not sure this is the best idea. I agree that there is a lot of stuff in this library which are not a good practice for c++11 but the trade-off was made in many places to ensure simplicity over performance, optimality. 

What do you think @owennewo ?

owennewo:
I'm slightly ambivalent, but probably agree with  @askuric. Adding it will help us (library developers) catch future refactoring bugs but will add mental load to library users many of who will be hobbyist arduino coders.

QiayuanLiao:
I agree with @cousinitt. 

Hobbyists could also end up writing a member function that overrides a base-class virtual when they didn't intend it.

askuric:
Ok guys, you thought me a nice c++ feature and changed my mind. Pull request accepted

PR #18, opened 2020-09-08
Refactored some logic in BLDCMotor.

cousinitt:
I *love* ternary statements, probably too much.
However I think that here a plain old if-else makes it more plain what is going on.


askuric:
I understand your concern. :D
But here I prefer to leave it inline. 

The reason is that this logic (ramping) is not important part of this function. And the expanded code is just shifts focus of the PID controller function towards this logic. 
I wanted the main functionality of the PID to be clearly written and easy to follow and understand. And then if someone wants to understand it in depth, for example the ramping part, well they can easily expand the code and see what it does exactly later.

cousinitt:
No sweat, I'll close the PR.

PR #26, opened 2020-10-02
Refactoring PID code

cousinitt:
I thought that it might be neat to add some smarts to the PID structure - to make it a functor. This would allow us to pull out some of the PID logic from the BLDCMotor class, and perhaps reuse it elsewhere as well.

If this is something that you think is an improvement, I could also do something similar with the low-pass filter.

Since this isn't really user-facing (i.e. in the "ino" file) it won't hurt/help understanding for most users of the library.

(as always - no hurt feelings if you don't like, I can close to clean up the PR part of your page).

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -21,14 +21,11 @@ BLDCMotor::BLDCMotor(int phA, int phB, int phC, int pp, int en)
 
   // Velocity loop config
   // PI controller constant
-  PID_velocity.P = DEF_PID_VEL_P;
-  PID_velocity.I = DEF_PID_VEL_I;
-  PID_velocity.D = DEF_PID_VEL_D;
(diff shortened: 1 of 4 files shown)


cousinitt:
I should either get/build one of the shields and try this out on an actual motor ... it isn't so much of a work in progress, but rather something that is adapted from something else I was working on.

Should I close this one as well (I didn't look at the new code you had for the library yet)?

askuric:
This one is included also yes. 
Both are in the dev branch

owennewo:
@cousinitt and @askuric - I've just looked at dev, you guys have been busy!

I like what i see. Thankyou both!

Looking forward to 1.6 release.

cousinitt:
Ah - neat, it is already in devel branch! I'll cancel this PR.

PR #27, opened 2020-10-02
Low pass filter refactor.

cousinitt:
Similar to the PID refactor, this pulls out the low pass filter code as a functor.

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -6,6 +6,12 @@
 // - cpr           - counts per rotation number (cpm=ppm*4)
 // - enable pin    - (optional input)
 BLDCMotor::BLDCMotor(int phA, int phB, int phC, int pp, int en)
+  : LPF_velocity(DEF_VEL_FILTER_Tf);
+  , PID_velocity(DEF_PID_VEL_P,
+                 DEF_PID_VEL_I,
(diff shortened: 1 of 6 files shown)


cousinitt:
Hmm ... I should have figured out a way to separate those.
If you actually like this I will close this and start over clean, otherwise, I'll close this one and trash all of it.

askuric:
Hey @cousinitt!
Nice work!

I wanted to do the same thing long time ago. It definitely makes more sense to have it this way than to put all of these things in a single file. I agree. 

I was a bit worried not to add too many files to be included. But I agree that at this point the library has grown and there is a lot of includes and therefore there is no need to procrastinate any more. 

I am working on a stepper motor implementation of the FOC and I will publish it today I hope, so let me see how can I integrate all of these changes into the new code and then I'll either accept the PR or just reuse the code maybe if eaiser. 

Thanks once again for digging deep into the library :D

askuric:
Hey @cousinitt, the new library release is in the dev branch. 
There I have included a good part of your code. 
I had to change some parts due to some simple errors and naming. 

Since there has been a lot of changes in the code before your PRs I could not merge them easily.

cousinitt:
Sounds great, glad that I could contribute. :)
I'll just close this PR then.

PR #28, opened 2020-10-12
adding support for two i2c buses + experimental checkBus() method

owennewo:
I think 1.6 has been released - this is not compatible with 1.6 as I've removed ability to set clock and sda/scl pins.  I can make it compatible but showing you here what I would/should have done.

Background: `TwoWire` implementation is vendor specific with small changes to contructor and begin method.  Also I now believe it's better to set things like clockspeed and sda/scl pin outside of sensor implementation (after alll the bus might be used by loads of other sensors).

The new i2c_dual_bus example shows how it would be used, but to summarise an optional TwoWire reference can be passed to `sensor.init(&Wire1)`.  Now that 1.6 has been released to make it backwards compatible would require me to add `clockspeed` and `sda_pin` and `scl_pin` back in.  I really don't mind doing this - it just makes the `MagneticSensorI2C.cpp` more tied to specific vendors.

### DIFF
--- a/examples/motion_control/position_motion_control/magnetic_sensor/angle_control/angle_control.ino
+++ b/examples/motion_control/position_motion_control/magnetic_sensor/angle_control/angle_control.ino
@@ -10,9 +10,9 @@
 #include <SimpleFOC.h>
 
 // magnetic sensor instance - SPI
-MagneticSensorSPI sensor = MagneticSensorSPI(10, 14, 0x3FFF);
+MagneticSensorSPI sensor = MagneticSensorSPI(AS5147_SPI, 10);
 // magnetic sensor instance - MagneticSensorI2C
-//MagneticSensorI2C sensor = MagneticSensorI2C(0x36, 12, 0x0E, 4);
+//MagneticSensorI2C sensor = MagneticSensorI2C(AS5600_I2C);
 // magnetic sensor instance - analog output
 // MagneticSensorAnalog sensor = MagneticSensorAnalog(A1, 14, 1020);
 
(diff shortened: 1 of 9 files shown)


askuric:
Hey Owen, 
I like this a lot. I think it is much better idea. 

Honestly, I wanted to implement this from the beginning, but I was struggling to implement `TwoWire` class into the magnetic sensor class for Arduino devices. Or maybe it was the `SPI` class. 
I did not spend too much time on it, though.

Have you tested this modification with Arduino devices?

I wanted to release this 1.6 version o the library a bit prematurely mainly because of steppers. So that the next version is a bit more robust. :D

We should implement this version of the bus handling for SPI and I2C for the next lib verison. And I would not bother too much with backward compatibility because this code will be much more clear and easy to use. :)

owennewo:
I've tested the new dual_i2c_bus example using 2x AS5600 on an esp32 and an stm32 (disco_f407 dev board).  Both work.  The stm32 took longer than it should because the second "i2c2" bus wasn't properly configured in [arduino_core_stm32](https://github.com/stm32duino/Arduino_Core_STM32/pull/1203).  I've not tested on atmega328p but I don't think that actually has 2x i2c buses.  We just need to test whether single i2c works (can't see why it won't).

I want to test a bit of SPI.  I don't think dual SPI bus support is as important as you don't get the chip address collision problem that you have with i2c.  I specifically want to make use of the SPI2 of my storm32 (the pins for SPI1 are connected to the driver chips).

owennewo:
I've added an experimental `checkBus()` method that clears `i2c` locks on my stm32 board.  I'd say about 15% of the time I upload or reset this board it happens when `AS5600` is holding `sda` low.  This cause sensor.init() to fail and you get no data from sensor.

This code checks whether `SDA` is being held low and simply sends 16 clock signals on the bus which from my experience is enough to convince my `AS5600` to release `sda`.
It should be called before init:
```
sensor.checkBus();
sensor.init();
```

PR #29, opened 2020-10-18
Initial support for secondary SPI bus

owennewo:
I've not tested 'dual' spi but have tested a single SPI using non default pins such as storm32 whose SPI1 pins are being used for motor drivers.  The SPI2 pins on the storm32 are sort of available after desoldering a couple of resistors and an LED.

The pattern is the same as i2c and I've added an example.


askuric:
I like it!

PR #32, opened 2020-10-22
fix spi datamask (word/int bug)

owennewo:
Note: This is targetting master (not dev) - I branched from master which is ahead of dev for some reason!

Hmm, fixing a bug I introduced.  The previous version had no right to work correctly at all (but did).

Looking at previous code 
```
const static word data_mask = ~(0 >> (16 - bit_resolution));
```
The intention was to start with 0 then do a ~ (bitwise not) to get b1111111111111111 (a word full of 1s) and then shift it (e.g. 14bit sensor yields mask of b0011111111111111 (14 1s in lowest bit of word)
However the zero decides to be 32bit int not 16bit word!

After assigning the int to a word it typically ends up being 16 bits of 1s - i.e using it has a mask has no effect.

Anyway weirdly the angle ends up being from roughly 18 to 18+2PI and the zero offset hides the 18 (and hides the bug).  So it worked without a working mask!

Anyway the new code is more readable and works:
```
const static word data_mask = 0xFFFF >> (16 - bit_resolution);
```
Update: Also noticed the angle_register for AS5047 is incorrect.  It was 0xCFFF but datasheet says 0x3FFF.  I have the AS5047u and both values seem to work but I suspect 0xCFFF is unknown and it returns the angle value as a default for unknown requests.


askuric:
Ok, let me merge this to the master and then lets have a chat one of the next days to go though this more thoroughly.

PR #39, opened 2020-12-13
SimpleFOC with OSC control example 1

runger1101001:
Here is a simple example for OSC control of SimpleFOC. I think most users will be able to follow this code, and take their own implementations from this point.

I think that, especially with MCUs like the ESP32, it is a really nice solution for controlling a motor, for example in a performance, art-installation, animatronics, laboratory, or just when testing.

It is nice to have a GUI for tablet/smart-phone, that does not need custom coding to set up the GUI side.


askuric:
hey Richard, 
I'm very happy to include it, thanks! :D

PR #51, opened 2021-01-31
print out how to skip calibration and disable motor on calibration fail

owennewo:
PR to

1. printout how to skip calibration for sensors where this is possible
2. disable driver if align fails

The reason this is a PR is that it might be using a little more "STRINGS" which might take more space on atmega328p.  So wanted to check whether this is worth the extra space.


askuric:
OK, I think most of it has been added to the v2.1, do you think there are still things to add to the new implementation?

PR #52, opened 2021-02-10
dual h-bridge v1.3 support

zjor:
Hi, 
I managed to make this stepper driver (https://wiki.dfrobot.com/MD1.3_2A_Dual_Motor_Controller_SKU_DRI0002) work with the following snippet
https://gist.github.com/zjor/a49b417d9904ab6dff6d3a2262b3fcff

See the demo on the video: https://www.youtube.com/watch?v=em70gQPN5b0

But when I tried `motor.move(3)` with velocity open loop, the movement was jerky, it didn't rotate properly.
Also I had to add `_delay(1);` to setPwm(..) which is also wrong.

### DIFF
--- a/src/drivers/StepperDriver2PWM2Pin.cpp
+++ b/src/drivers/StepperDriver2PWM2Pin.cpp
@@ -1,0 +1,88 @@
+#include "StepperDriver2PWM2Pin.h"
+
+StepperDriver2PWM2Pin::StepperDriver2PWM2Pin(int ph1PWM, int ph1Dir,int ph2PWM, int ph2Dir, int en1, int en2) {
+  // Pin initialization
+  pwm1 = ph1PWM; //!< phase 1 pwm pin number
+  dir1 = ph1Dir; //!< phase 1 dir pin number
+  pwm2 = ph2PWM; //!< phase 2 pwm pin number
+  dir2 = ph2Dir; //!< phase 2 dir pin number
+
+  // enable_pin pin
+  enable_pin1 = en1;
+  enable_pin2 = en2;
+
+  // default power-supply value
+  voltage_power_supply = DEF_POWER_SUPPLY;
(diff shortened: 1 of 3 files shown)


src/drivers/StepperDriver2PWM2Pin.cpp
```diff
@@ -0,0 +1,88 @@
...
+  Ualpha = _constrain(Ualpha, -voltage_limit, voltage_limit);
+  Ubeta = _constrain(Ubeta, -voltage_limit, voltage_limit);
+  // hardware specific writing
+  duty_cycle1 = _constrain(abs(Ualpha)/voltage_power_supply,0.0,1.0);
+  duty_cycle2 = _constrain(abs(Ubeta)/voltage_power_supply,0.0,1.0);
+  // set direction
+  digitalWrite(dir2, Ualpha > 0 ? HIGH : LOW);
+  digitalWrite(dir1, Ubeta > 0 ? HIGH : LOW);
+
+  _delay(1);
```
zjor, line 77: this is bad, I know, but otherwise, the movement was not smooth

askuric:
Hey @zjor ,
I could not commit to your branch so I just merged your code directly to the dev branch. 
At the end I deleted the new class that you created, just because it is basically the same as the one we already have `StepperDriver2PWM` so I just created a new constructor which takes the same arguments that you described in your code. 
` StepperDriver2PWM(int pwm1, int dir1, int pwm2, int dir2, int en1, int en2) `

Now regarding `_delay(1)` :D
I have removed it because even if it is needed this is not the place to do it. You should probably do it in the Arduino loop function. 

Could you please check if the new code works for you? 

And I'd like to discuss the reasons why you need this _delay(1) in your code in the community thread:
https://community.simplefoc.com/t/stepper-motor-setup/519/3

I have some ideas why and maybe we can avoid it :D

PR #53, opened 2021-02-20
SAMD21

runger1101001:
Hi Antun,

Here the first pull request for SAMD support.

Please take a look, let me know what should be changed. At the moment it is against master - should I do it against the dev branch instead?

Regards from Vienna!


askuric:
This is awesome!
But I have a bit of trouble going through the changes since the OSC examples are also induced in changes. Could you please rebase it to the dev branch and we'll merge the osc example separately so that the commits and pull-requests are easier to follow.  
Great work!

PR #55, opened 2021-02-21
remove extra pin for 3 pwm config on teensy 

mc18g13:
remove extra pin for 3 pwm config on teensy and align 4pwm config with header


askuric:
Hey @mc18g13!
thanks for the pull-request. These changes have already been included into the dev branch. :D
We will release it very soon. (we need to document it first)

Teensy has had a compilation error because I actually do not have on, and I forgot to compile test compile it :D
Thanks for going through the code so thoroughly!

Would you be interested to try to implement 6pwm mode maybe?
Otherwise I'll close the pull-request.

mc18g13:
Ah!! of course... Check the dev branch. I'd of thought this could be setup with the CI? I don't mind looking into adding teensyduino into the workflow

I am actually porting some of this work to c for Raspberry Pico project I am working on. Unfortunately not looking very Arduino like or back portable but I think the pico libs are going to be ported to Arduino soon.

Any glaringly obvious issues I am missing about adding the 6pwm? Not all the pins are supported for pwm on some of the older versions of teensies and I don't have the kit to actually test the library with it. But I can plonk it on a scope?

askuric:
Hey @mc18g13,
Sorry about the delay. Have you had some time to try to do some 6pwm implementation? 
I would be very happy to help if you are in to spend some time on it :D

Its not too complicated but it will be a bit more difficult than the rest because we really need to dig into the Teensy timers.

PR #56, opened 2021-02-23
Samd21 dev

runger1101001:
Here is the SAMD21 support updated against the dev branch. I tested it on an Nano 33 IoT, and could not see any difference to the version that I'd branched from master.

I had to make one change to the new Commander class: it was using "HardwareSerial", but SAMD21 "standard serial" is via the USB port, and uses "USBSerial". So I changed Commander to user "Stream", which is the common subclass of both of these. The code itself was not using any methods that aren't in Stream anyway.

This looks better now, and could be merged without conflicts.

The only changes to existing code are the mentioned Commander class changes, and adding the SAMD21 define in generic_mcu.cpp.

### DIFF
--- a/examples/hardware_specific_examples/SAMD_examples/README.md
+++ b/examples/hardware_specific_examples/SAMD_examples/README.md
@@ -1,0 +1,62 @@
+
+# SAMD Support
+
+SimpleFOC supports many SAMD21 MCUs, really any SAMD21 supported by Arduino core should work.
+
+## Pin assignments
+
+The SAMD chips have some very powerful PWM features, but do not have flexible pin assignments.
+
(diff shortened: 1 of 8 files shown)


askuric:
Richard that is perfect!

I have a question before I merge it. Are you sure that `Stream` is the same as `HadrwareSerial`?
I was using `Print` before, but in `Commander` I really need to use `Serial.available()` and so on...

I suppose we can create different types of handlers for different communications.
But for now, I might leave it as `HardwareSerial`, just for the sake of finishing this huge release :D

runger1101001:
Awesome! I will do some tests on the merged code and report back.

PR #57, opened 2021-02-23
Comprehensive OSC example on ESP32

runger1101001:
Second, more comprehensive OSC based control example.

Includes:

Initialization of 2 3-PWM motors

-  with I2C magnetic encoders
- Initialization code for OSC via UDP on ESP32 WiFi
- ESP32 OTA (program your ESP32 over the WiFi)

Example pureData GUI & Control for 2 motors, including parameter tuning of all the parameters, and various control options.

- Voltage, Velocity & Position Control
- Live display of Set-Point and angle
- Control motors individually or together


askuric:
Very cool, I'm going to merge it!

We have to talk a bit about these examples, I was thinking of using for my balancer. To make the control app integrated on the esp :D

PR #58, opened 2021-02-23
SPI driver for TI's DRV8316 motor IC. Supports all documented registers

runger1101001:
SPI driver for the DRV8316. You need to communicate with it via SPI to configure the chip. It's essential if you want to run 3-PWM instead of the default 6-PWM.

Once you have SPI configuration working, you can do all kinds of things, like switch on and off the various protections, control the voltage of the built-in buck-regulator, etc...

This driver supports all registers and functions that are currently documented by TI.

The driver has a human-readable API, and defined Enums for the various configuration options. This means the user should be able to configure the driver IC without referring back to the datasheet all the time.

I created a sub-directory within "drivers" for these drv8316 files, since there are 3 of them. I didn't put them in the "hardware-specific" folder, since this so far contains only the mcu-specific implementations of the commutation code, and it didn't seem right to mix in driver-chip specific code there. After all, the DRV8316 can be used with any MCU.


askuric:
Very cool, very systematic approach, great work. I am not really sure what is the best way to integrate this, let's discuss this one with the team. 

Maybe we would need to figure out a way to make an additional library of drivers/sensors/....

runger1101001:
This will be moved into it's own library, as discussed. I've set this up, and am just trying it out. It shall be called: "Simple FOC drivers"

PR #60, opened 2021-03-15
Removed encoder init from mag sensor sketch

armaanhammer:
Removed extraneous encoder initialization from magnetic sensor pole pair finder utility


askuric:
Thanks a lot!

PR #61, opened 2021-03-18
adding MagneticSensorPWM implementation

jordancormack:
Adding support for PWM magnetic position sensors, tested with the AS5048a's PWM output. This implementation is almost identical to the already existing MagneticSensorAnalog sensor type, giving an output of 0 to 2Pi radians per mechanical revolution.


askuric:
Hey @jordancormack 
Thanks for the pull-request. I'm sorry for a bit of delay. I am going to merge it to the dev branch but I dont think that this implementation will be the one we will use. I'll use your code to bootstrap the pwm sensor implementation, but we will be uisng an interrupt based support. 
`pulseln` is blocking function and it will take too much time to execute. Especially since the pwm sensors have pwm frequency of 1kHz which means that `pulseln` takes at least 1ms to execute. Which is ok to test stuff, but it is too long for smooth operation. 

Thanks!

PR #64, opened 2021-04-06
Code Style: format "tab" to "two space"

imliubo:
I found that the tab and space are mixed in the code, resulting in some code formats that are not very uniform. According to observations, most of the indentation is two spaces, so I replaced some tabs with two spaces.

example: [here](https://github.com/simplefoc/Arduino-FOC/blob/master/src/StepperMotor.h#L44)

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -16,7 +16,7 @@ BLDCMotor::BLDCMotor(int pp, float _R)
 
 
 /**
-	Link the driver which controls the motor
+  Link the driver which controls the motor
 */
 void BLDCMotor::linkDriver(BLDCDriver* _driver) {
   driver = _driver;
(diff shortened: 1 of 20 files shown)


askuric:
Hey @imliubo, 
Thanks for the time to look into the code, but this pull  request changes 20+ files and thousands of lines of code. And changing this many files I am not comfortable of doing without proper testing. 

Now when it comes to convenience changes such as tab/space convention it is up to each one of us to respect it, but I agree that it is sometimes annoying when it happens that the files are not well formatted. But at this moment, I cannot guarantee you that it will not happen again even if I merge your pull request. So for now I'd leave it as is. 
I hope you understand :D

imliubo:
@askuric ，
Got it.Perhaps a certain code specification can be implemented from now on.

PR #69, opened 2021-04-29
random fixes related to current sensing

sDessens:
Current control seems to work as expected with these changes on the B-G431B-ESC1.


askuric:
Awesome, thanks for going into such a detail!
I'm merging!

PR #72, opened 2021-05-02
Dev

runger1101001:
3 changes:

1. added an enable_active_high field to BLDCDriver6PWM. The DRV8316 has an "drvoff" pin - e.g. an enable pin that is active-low. This simple change enables configuring the enable/disable function to support active-low.

2. SAMD21/SAMD51 PWM code updates. SAME51 support (trivially, code is same as for SAMD51).

3. moved the location of the sensor calls in move() and loopFOC() to before the enable checks. This means SimpleFOCStudio keeps displaying the values, and also disabled motors don't lose track of full rotations.


askuric:
Awesome, great work! 😀

Issue #46, opened 2021-01-02
motor.command() and CRLF


owennewo:
This confused me - so I'm raising an issue for discussion.

`motor.command(String)` doesn't work well with Serial set to eol=CRLF (i.e. `\r\n`)

With CRLF all commands are 'sets' e.g. `P` will set the P term to zero.

A fix for this is to either configure your serial terminal to LF (`\n`)
e.g. platformio (which defaults to CRLF)  you could set the following in you platformio.ini
```
monitor_flags =
  --eol
    LF
```

If we wanted to be more 'tolerant' we could change the code e.g. to be:
```
  char GET = user_command.charAt(1) == '\n' || user_command.charAt(1) == '\r';
```

askuric:
https://community.simplefoc.com/t/node-oriented-communication-interface-for-the-simplefoc/466/14

askuric:
You're right. If you have time go for it. :)

askuric:
https://community.simplefoc.com/t/node-oriented-communication-interface-for-the-simplefoc/466/15

PR #75, opened 2021-05-10
Issue #46: motor.command() and CRLF

maxlem:
The eol character is now configurable
The user can now actuvate echo feedback to see what he is typing
I also removed a duplication in two run() overload, while fixing what I think was a mistake by enforcing the use of the provided argument Stream& even if the ctor one is not null.
Lastly, I warn the user if \n is configured but \r is detected (only in user friendly mode)
All the past code should work as before (no API change required)

### DIFF
--- a/src/communication/Commander.cpp
+++ b/src/communication/Commander.cpp
@@ -1,11 +1,14 @@
 #include "Commander.h"
 
 
-Commander::Commander(Stream& serial){
+Commander::Commander(Stream& serial, char eol, bool echo){
   com_port = &serial;
+  this->eol = eol;
+  this->echo = echo;
 }
-Commander::Commander(){
-  // do nothing
+Commander::Commander(char eol, bool echo){
+  this->eol = eol;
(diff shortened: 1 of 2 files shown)


askuric:
I'm ok with this PR, I'm merging it now. 

> The user can now actuvate echo feedback to see what he is typing

I'll leave it for now, but I'd like to solve this a bit more consistently with the library API in future. Introduce a `VerboseMode::echo` or something like that. 

> I also removed a duplication in two run() overload, while fixing what I think was a mistake by enforcing the use of the provided argument Stream& even if the ctor one is not null.

This was a feature not a bug :D
So that the user can use more than one serial if he wants or that he does not have to define serial in the constructor. But there are problems with this implementation, I agree. Like the common buffer and similar. 

 > Lastly, I warn the user if \n is configured but \r is detected (only in user friendly mode)
 
 This is a good idea, but the text is too long for Arudino UNO. I'll shrink it a bit :D

PR #77, opened 2021-05-10
Dev

runger1101001:
Simple fix to prevent buffer overruns (and crashes) if other side sends commands which are too long.


askuric:
I'm in!

PR #79, opened 2021-05-12
Dev

maxlem:
this contains my other MR which askuric says he's fine with it

https://github.com/simplefoc/Arduino-FOC/pull/75

let me know if you want me to split them in two


askuric:
No need, the other part of the code is merged already. 

@runger1101001 are you ok with this serial modification?
Should we leave the `Serial` class usage by default?

runger1101001:
Basically the SAMD stuff is fine, but there are a lot of changes in Commander that I am uncertain about... I think unfortunately you guys were working on that in parallel.
I'd love to include the SAMD changes though, that makes total sense. I already did the Pico code's debug output in the same way...

At this point the way I would do it is to merge the changes over manually (locally), so that I can exclude the commander changes. I am more than happy to do this, but if you @maxlem  would prefer to send another pull request excluding the commander changes, then GitHub would track that correctly. If I do it locally it will look like the change came from me...

runger1101001 (changes requested):
Basically perfect, except we have parallel changes in Commander.cpp which can't be merged. Without the commander changes I would merge it!

maxlem:
Well, no more conflicts, but it's a mess.

I simply forgot to create a branch when I started to work on ADC-DMA stuff, it was a different feature.

In a perfect world, I would scrap all this and start a new fork, cherry-picking all the changes I really want to make it into dev

I'm so sorry :(

Tell me what you'd like I'll do it.

We cand always leave the ADC/LowSide stuff as it is since we're in a dev branch, but while the ADC (current_sense/hardware_specific/samd21_mcu.h) class is in a usable state, the LowSideCurrentSense.h is plain wrong, returns volts instead of amps right now and I know @askuric is working on its own

runger1101001:
Wow, now it says 51 files changed 😱

So I checked it out locally and applied it to the clean dev branch, and it changes only 10 files, and only the SAMD-related stuff. It is a FF merge.

So I will merge it. It can be reversed if it turns out different online than when I do it locally...

runger1101001:
Perfect, it worked out on GitHub just the same way as locally.
Supernice! I will try to test it out in the next few days!

Thank you so much, @maxlem  for your hard work!

If I make make a suggestion, please rename your current "dev" branch (e.g. "dev-old") and create a new "dev" branch from the current simplefoc:dev branch. In this way your (future) changes will be easy to include.

PR #85, opened 2021-05-30
Use float instead of double for constants to improve speed

KaSroka:
We should use float constants instead of double (like `0.0123f`) to improve claculation speed. If we have double constant all expression will be calculated as double just to convert it to float at the end.

I have created a simple benchmark:
```c++
#include <SimpleFOC.h>

#define N 5000UL
#define BENCH(foo, real_foo, n) bench(foo, real_foo, n, #foo)

static float bench_calibration = 0.0f;

float empty(float x)
{
    return 0.0f;
}

float bench_time(float (*foo)(float), size_t n)
{
    uint32_t begin = micros();

    for (size_t i = 0; i < n; i++)
    {
        foo((float)i / n * _2PI);
    }

    uint32_t end = micros();

    return (float)(end - begin) / n - bench_calibration;
}

float bench_error(float (*foo)(float), float (*real_foo)(float), size_t n)
{
    float sum_rms = 0.0f;
    
    for(size_t i = 0; i < n; i++)
    {
        float angle = (float)i / n * _2PI;
        float real = real_foo(angle);
        float calc = foo(angle);
        float diff = calc - real;
        sum_rms += diff * diff;
    }

    return sqrtf(sum_rms / n);
}

void bench(float (*foo)(float), float (*real_foo)(float), size_t n, const char * name)
{

    Serial.print("Testing ");
    Serial.print(name);
    Serial.print(" ");
    Serial.print(n);
    Serial.println(" times");

    float time_spent_us = bench_time(foo, n);
    float rms = bench_error(foo, real_foo, n);

    Serial.print("Time: ");
    Serial.print(time_spent_us, 3);
    Serial.print("us, RMS error: ");
    Serial.println(rms, 6);
}

void bench_calibrate(size_t n)
{
    bench_calibration = 0.0f;
    bench_calibration = bench_time(empty, n);
}

void setup() {
    Serial.begin(115200);

    bench_calibrate(N);

    BENCH(_sin, sinf, N);
    BENCH(_cos, cosf, N);
}

void loop() {

}
```

On STM32F0 running at 48MHz it gave me results presented below.
Before change to floats, using doubles as constants:
```bash
Testing _sin 5000 times
Time: 34.277us, RMS error: 0.001612
Testing _cos 5000 times
Time: 46.258us, RMS error: 0.001612
```

After change to floats:
```bash
Testing _sin 5000 times
Time: 18.794us, RMS error: 0.001612
Testing _cos 5000 times
Time: 24.926us, RMS error: 0.001612
```

As you can see it's almost two times faster and we are talking about 32bit MCU, on 8bit Arduino it should improve even more.

I haven't touched MCU specific files but I can see some double constants here as well.

### DIFF
--- a/examples/hardware_specific_examples/Bluepill_examples/encoder/bluepill_position_control/bluepill_position_control.ino
+++ b/examples/hardware_specific_examples/Bluepill_examples/encoder/bluepill_position_control/bluepill_position_control.ino
@@ -1,9 +1,9 @@
 /**
- * 
+ *
  * STM32 Bluepill position motion control example with encoder
- * 
+ *
  * The same example can be ran with any STM32 board - just make sure that put right pin numbers.
- * 
+ *
  */
 #include <SimpleFOC.h>
 
@@ -32,10 +32,10 @@ void doTarget(char* cmd) { command.scalar(&target_angle, cmd); }
 
 
 void setup() {
-  
+
   // initialize encoder sensor hardware
   encoder.init();
-  encoder.enableInterrupts(doA, doB, doI); 
+  encoder.enableInterrupts(doA, doB, doI);
   // link the motor to the sensor
   motor.linkSensor(&encoder);
 
@@ -54,20 +54,20 @@ void setup() {
   // set motion control loop to be used
   motor.controller = MotionControlType::velocity;
 
-  // contoller configuration 
+  // contoller configuration
   // default parameters in defaults.h
 
   // velocity PI controller parameters
-  motor.PID_velocity.P = 0.2;
+  motor.PID_velocity.P = 0.2f;
   motor.PID_velocity.I = 20;
   // default voltage_power_supply
   motor.voltage_limit = 6;
   // jerk control using voltage voltage ramp
   // default value is 300 volts per sec  ~ 0.3V per millisecond
   motor.PID_velocity.output_ramp = 1000;
- 
+
   // velocity low pass filtering time constant
-  motor.LPF_velocity.Tf = 0.01;
+  motor.LPF_velocity.Tf = 0.01f;
 
   // angle P controller
   motor.P_angle.P = 20;
@@ -75,11 +75,11 @@ void setup() {
   motor.velocity_limit = 4;
 
 
-  // use monitoring with serial 
+  // use monitoring with serial
   Serial.begin(115200);
   // comment out if not needed
   motor.useMonitoring(Serial);
-  
+
   // initialize motor
   motor.init();
   // align encoder and start FOC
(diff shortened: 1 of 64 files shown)


runger1101001:
This is a nice idea! And nice test. May I ask, did you also test the changed code with a motor?

I would accept this change, but not against the master branch. Would you be prepared to send a pull request against the dev branch? If not it is not a problem, I can merge your changes over locally to the dev branch, but that would lose the "connection" to your pull request...

Thanks a lot!

KaSroka:
Hi, yes I will modify the PR to target the dev branch.

I have tested this with stepper motor and it allowed me to achieve better max RPM (600RPM increased to 765RPM)

KaSroka:
I have changed the base branch and included all the changes for the MCU specific code as well.

langwadt:
-fsingle-precision-constant -Wdouble-promotion

on AVR it shouldn't make any difference because doubles are defined as floats

KaSroka:
I agree with `-fsingle-precision-constant` but this is something we cannot enforce here, so it's better to sue correct types from the beginning. It will also make huge difference on MCUs with FPU when compiled without `-fsingle-precision-constant`.

I know this touches a lot of files so it may need a lot of testing.

langwadt:
yeh, if you can't add compiler options that's out, and it would only deal with constants, it won't replace things like sin() with sinf() if any of those functions are used

runger1101001:
https://mcuoneclipse.com/2019/03/29/be-aware-floating-point-operations-on-arm-cortex-m4f/

runger1101001:
So on SAMD21 I'm getting about 19% improvement on my loopFOC()...

I think I'm in favour of merging this, even though it touches so many files. I was able to apply it cleanly to my local branch.

askuric:
That's great to hear, well done @KaSroka  for this pull request.

I'm gonna test it today as well and we will merge it probably as is. I just need to make sure this does not break anything for hardware specific implementations. 

I actually think that these changes will impact the least the Arduino, because as far as I know they do not have double at all as a type. 

BTW. we do not use sine functions `sin` or `sinf` in simplefoc. We've got much faster array based implementation `_sin`. 
Also for square root we use the fast inverse square root.

askuric:
So the results of the testing say:
SimpleFOC PowerShield + Gimbal motor + encoder:
All the times are given in milliseconds and averaged over 1000 `move()` + `loopFOC()` calls.

MCU | `voltage` before| `voltage` after | `foc current` before |`foc current` after  | % gain
--- | ---- | ---| ---- | ---- | ---
ESP32 (r1 d32) | 33-45 us | 26-30us | 79-105 us| 66-82us |  ~30%
STM32 (f411re) | 79-86 us | 50 us | 243-257 us | 177 - 200 us | ~30%
SAMD51 (Metro m4) | 50-55 us | 40-45 us | 143-150 us | 104-143 us | ~20%
Arduino DUE|  174-220us | 160-195 us |  360-430us | 330-400 us | ~10%
Arduino UNO | 1000-1700us | same | 2100-3000us | same | ~0%

So overall the gain is pretty high, so we'll definitely try to make this a standard :D

Issue #82, opened 2021-05-26
MagneticSensorSPI and STM32


runger1101001:
As reported here:
https://community.simplefoc.com/t/sensor-as5047p-spi-reads-0-angle-with-stm32f103-aka-blue-pill/886/18

but also by other users, I think. It explains the difficulties some people (including myself) were having with magnetic sensors and STM32.

Here is my summary from that post:

Hey, I did a quick test setup:
[image]

- SimpleFOC 2.1.1 master branch (current normal version of the library)
- BluePill STM32F103C8
- AS5047P sensor
- PlatformIO, STM32 support v13.0
- Motor not connected, just the sensor

I can report the following:
It actually works out of the box, with no modifications to MagneticSensorSPI.cpp.
The resulting speed is too fast for my crappy logic analyser, the oscilloscope shows the clock line is at 9MHz:
[image]
That makes sense since we set the pre-scaler to 8, and 72MHz system clock / 8 = 9MHz SPI clock in this case.

Surprisingly, this is working fine even via the SPI hub and all the cables:
[image]
With my suggested modifications to MagneticSensorSPI, the picture changes to:
[image]
So now we don’t set the pre-scaler, but rather allow beginTransaction() to apply the clock_speed as specified in the settings object.
So we get 1MHz SPI clock, as specified in the settings.

To me this confirms the following:

the standard settings kind of work with STM32. But the bus speed does not get set, via the settings, resulting in a MCU-dependent default bus speed via the pre-scaler setting of 8, which is often much too high. I’m surprised my setup works at 9MHz, and I think if there were a BLDC driver and motor running chances are it wouldn’t work well any more.

the current code prevents the bus speed from being set correctly on STM32, so as a result, many people will have trouble with the current code under STM32 (as we see in this thread).

askuric:
Is this still an issue @runger1101001 ?

runger1101001:
No, this is fixed by removing the pre-scaler settings. I'm not sure where that came from originally. I suspect it was probably required in an older version of the STM32 platform, but is not required any more. Setting the SPI bus speed to the value given in the settings is definitely the better approach, IMHO.
So I think it can be closed.

PR #87, opened 2021-05-31
Stm32 sensor fix

runger1101001:
Fix for issue #82

This pull request modifies MagneticSensorSPI.cpp in the following ways:

1. Remove #ifndef STM32 masking out spi.beginTransaction and spi.endTransaction on STM32 MCUs. I can find no reason for doing this, and on the other hand doing so means the SPI settings don't get applied for the SPI call, meaning firstly that SPI clock speeds were all wrong for all STM32 users, and some users had clock speeds faster than the sensors they were using supported. Additionally there would be big problems using multiple SPI devices on the bus.

2. Remove the setClockDivider call - is there any architecture that needs it? Its an obsolete way of setting the SPI clock speed, the modern way is to use SPISettings and beginTransaction(). @askuric do you know where this came from?

3. Change the 10us delay, between calls, replacing it with a 1us delay. The sensors I can find (AMS, MagAlpha) all require significantly <1us downtime between requests, so this should be sufficient. For ESP32 there is a 50us delay (!) - anyone know why? I've left this for the moment.

### DIFF
--- a/src/sensors/MagneticSensorSPI.cpp
+++ b/src/sensors/MagneticSensorSPI.cpp
@@ -75,11 +75,8 @@ void MagneticSensorSPI::init(SPIClass* _spi){
   
 	//SPI has an internal SPI-device counter, it is possible to call "begin()" from different devices
 	spi->begin();
-#ifndef ESP_H // if not ESP32 board
-	spi->setBitOrder(MSBFIRST); // Set the SPI_1 bit order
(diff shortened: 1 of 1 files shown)


runger1101001:
I think this can be merged into dev branch, so it gets some wider testing...
It changes only one file, and can be easily backed out if it causes problems. It's working for me.

Issue #84, opened 2021-05-30
Add code format checking


KaSroka:
This repo would benefit from keeping the code formatting more strict. Especially spaces around operators are all over the place.

What about adding https://github.com/marketplace/actions/clang-format-check github action?

runger1101001:
Hi @KaSroka 

Thanks for the suggestion, and the link. I'm of two minds:

1. I've never seen focusing on syntax actually really produce any tangible benefits. And personally, with limited time to spend on the project, I'd rather fix actual bugs or code cool new functionality. In terms of other users, I think they would benefit more if the time were spent on documentation rather than syntactic changes.

2. As a developer and computer scientist, there is a side to me that likes order, and I do try to keep the code I write "tidy" to my eyes. So in theory it would be nice if the code were all well formatted.

Ultimately I see this as Antun's call. Personally the 1. reasons would outweigh the 2. reason for me, and I wouldn't want to reformat all the code myself.

KaSroka:
@runger1101001 Reformatting is handled by clang-format when necessary.

Regarding 1. as a professional C programmer I can tell you that strict code formatting is a key to code maintainability. If it's ok I can create a PR with proposed changes so you can review it.

PR #88, opened 2021-05-31
Add clang format

KaSroka:
Add clang-format support as discussed in #84.

This includes code reformatting to conform with proposed formatting as well as github action to check this. The only downside is that this action doesn't check `.ino` files, I haven't figured out how to check them.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,126 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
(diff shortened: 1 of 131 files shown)


runger1101001:
Hey, that's cool, thank you for this work!

I must leave it to @askuric  to decide this - it changes *all* the files. Normally we don't accept changes that modify too much of the code, but perhaps this is a special case.

Just so you understand our hesitancy, please keep in mind that sweeping changes to the code mean someone has to test things. And we support:

- ATMega
- ESP32
- STM32
- RP2040
- SAMD21/51
- Teensy
- Due

- SPI Sensors
- I2C Sensors
- Enoders
- Hall Sensors
- Current Sensing

- BLDC Motors
- Stepper Motors

So sweeping changes mean someone has to physically test a very large number of hardware combinations to ensure we haven't broken anything... for this reason we need to be a bit careful.

But we really appreciate the effort you are putting in! Thank you!!

KaSroka:
I fully understand it! It's always up to maintainers to accept or decline PR.

KaSroka:
Also from testing perspective - it's not changing logic in code but only formatting so compilation check + review looking for unnecessary changes should be sufficient.

askuric:
Hey @KaSroka, as you can see we are very hesitant to merge this kind of PRs to the main library. There are many changes and we are do not have a specific formatting style that we want to settle for the moment. :D
Thanks for the proposition, but I prefer not to include it to the library at the moment.

PR #91, opened 2021-06-06
Sensor fixes

runger1101001:
I'm fairly happy with this change... I'd like it to get tested more widely, to see if other architectures have problems...

I don't know when you plan the next release, but if there is still a little time it would be cool to merge this and get it tested.

The patch modifies the following:
- new Sensor base class API, simplifying the implementation of new sensors to a single method: getSensorAngle()
- Sensor base class stores shaft angle (0..2PI) as a float, and full rotations as a uint32_t
- Sensor base class updates (calls getSensorAngle() and updates internal fields) on calls to updateSensor()
- getVelocity() is handled in base class (but can be overridden if desired)
- all the existing sensor classes have been updated to the new API

So far MagneticSensorSPI has been tested on SAMD21 and STM32. Overall, there is a slight performance improvement since the number of sensor polls is reduced by 1.

In addition I have tried to identify all the other precision problems in BLDCMotor.cpp, and added comments. I don't think these problems are as severe as the velocity problem (which users are often finding), as they are more the kind of problem that occurs if you're tying to adjust the position very precisely after having gone through a few thousand rotations. I don't know how often that happens in real use.

### DIFF
--- a/examples/hardware_specific_examples/Bluepill_examples/encoder/bluepill_position_control/bluepill_position_control.ino
+++ b/examples/hardware_specific_examples/Bluepill_examples/encoder/bluepill_position_control/bluepill_position_control.ino
@@ -1,9 +1,9 @@
 /**
- * 
+ *
  * STM32 Bluepill position motion control example with encoder
- * 
+ *
  * The same example can be ran with any STM32 board - just make sure that put right pin numbers.
- * 
+ *
  */
 #include <SimpleFOC.h>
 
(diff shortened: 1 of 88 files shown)


dreusel:
I think this change leads to an issue where getVelocity() for I2C magnetic sensors returns a valid value only the first time it is called in a FOC loop, because it uses and then changes the _prev class variables. After that it will just return 0. 
Perhaps the calculated velocity could also be kept as a class variable and re-used until update() is called.

Awesome work btw. I ran into this issue when I was doing some experiments with angle-correction based on current velocity, which was already a lot easier to experiment on with the work from this PR in place.

PR #103, opened 2021-07-23
Commander.cpp: fix 

IhorNehrutsa:
If call_list[i](&user_input[0]) 
Then request from SimpleFOCStudio will work.
https://github.com/JorgeMaker/SimpleFOCStudio/blob/0e460b665964308d3e6dba8d61d5473a58c26b2b/src/simpleFOCConnector.py#L608-L624
```
 def updateStates(self):
        if self.isConnected:
            self.getCommand('MG0')
            time.sleep(100 / 1000)
            self.getCommand('MG1')
            time.sleep(100 / 1000)
            self.getCommand('MG2')
            time.sleep(100 / 1000)
            self.getCommand('MG3')
            time.sleep(100 / 1000)
            self.getCommand('MG4')
            time.sleep(100 / 1000)
            self.getCommand('MG5')
            time.sleep(100 / 1000)
            self.getCommand('MG6')
            time.sleep(100 / 1000)
```
I am not sure about this. Please review.

### DIFF
--- a/src/communication/Commander.cpp
+++ b/src/communication/Commander.cpp
@@ -90,7 +90,7 @@ void Commander::run(char* user_input){
     default:
       for(int i=0; i < call_count; i++){
         if(id == call_ids[i]){
-          call_list[i](&user_input[1]);
+          call_list[i](&user_input[0]);
           break;
         }
       }


askuric:
Hey @IhorNehrutsa ,

This should not work. The command is always the first character of the user command and the rest of the characters is sent to the callback functions that are subscribed to the particular letter. For example if you send a command `ME0`, character `M` will be the command id of a motor for example and in the motor callback you will send the `E0` not the whole command `ME0`. So the code is correct, it should remain: 
```
call_list[i](&user_input[1]);
```

the state update function of the simplefocstudio should be written in a bit better manner, because at the moment it sometimes tries to reach the microcontroller too early (when it is intializing) and this causes the timeout errors and the whole state update stops. 
Maybe this is the behaviour that you've seen. 
But this does not have anything to do with the commander. :D

IhorNehrutsa:
I tryed to use "MMGn" command in SimpleFOCStudio.
```
 def updateStates(self):
        if self.isConnected:
            self.getCommand('MMG0')
            time.sleep(100 / 1000)
            self.getCommand('MMG1')
            .........
```
It works, but It looks not clear.

askuric:
I'll close this PR for now and we can talk continue the discussion in the issue.

PR #108, opened 2021-08-21
Create atmega32u4_mcu.cpp

candyriver:
add support for atmega32u4.
just modify the _pinHighFrequency function.


askuric:
Hey @candyriver,

I'm going to merge your code to the atmega328_mcu.cpp with some compiler flags so that we avoid repeating too much code. 
Thanks for your contribution!

Issue #121, opened 2021-10-20
No sensor class calls Sensor::init()


steph643:
For example `MagneticSensorSPI` never calls `Sensor::init()`. I believe this causes `Sensor::full_rotations` to be wrongly computed during the first call to `Sensor::update()`.

askuric:
Hey @steph643,
The sensor init is delegated to the user. The user has to call it before linking the sensor to the motor.
I am not sure if that makes sense in your case.

Do you still have the same issue with miscalculated values even after the  init call?

runger1101001:
This is my bad - I can't believe I missed this! I fear it is because I normally use the specific sensor implementations I have created in the Arduino-FOC-drivers repository. Those all call the base class init method... doh! silly and careless me!

I think Encoder and HallSensor don't need it, since their velocity calculations are handled in the specific classes overriding the base class. The other sensors all need it.

I have fixed it in my fork: https://github.com/runger1101001/Arduino-FOC/tree/dev
and created a pull request against the dev branch. I won't merge it just yet because I'm on vacation and can't test with a real sensor and motor.

PR #127, opened 2021-10-30
Fix for #121 - missing Sensor::init() calls and NRF52 support by Polyphe

runger1101001:


askuric:
Hey Richard, 
I've tested the code and I'd like to merge it, I've got just one question.  Should this samd code be merged also?

runger1101001:
Yes, I think it can be merged. It makes the AREF pin optional in SAMD current sensing, so you can use the internal reference if you can't spare the AREF pin or don't want to use it.
It should be ok to merge, the SAMD current sensing isn't quite working for me yet anyway, so I don't think it makes things worse :-)

Sorry it got mixed in with the other changes, but it was already on my dev branch when I merged the latest release to my branch...

PR #134, opened 2021-11-12
SAMD code improvements

runger1101001:
* The PIN assignment issue on SAMD/E51 has been fixed
* Removed unnecessary syncs which were slowing down setting the PWM speed in some cases
* SAMD/E51 now uses the DPLL clock source, for faster PWM frequencies
* pwm_frequency can now be set for both SAMD/E51 and SAMD21
* on SAMD51 it should be possible to set it between 2kHz and 150kHz


runger1101001:
Since this affects only the SAMD files, I will merge it to the dev branch.

runger1101001:
I have tested so far with:
- Metro M4
- Feather M4 CAN
- Arduino MKR1000

I'll test a few more over the next couple of days.

PR #126, opened 2021-10-30
Add new MCU support "NRF52 series"

Polyphe:
I added PWM drivers for the NRF52 family. I tried the NRF52832 and the Micro:bit V2 (NRF52833), and that works very well. 

That should work on NRF52840 like for example Nano BLE33.

This library directly uses the PWM registers of the NRF52, not the NRF library (nrfx_pwm.h)

PR #137, opened 2021-11-17
fix nRF52 support to compile for Nano 33 (mbed)

runger1101001:
Also see comments under PR #126 

Makes the nRF52 support compile on the Arduino boards (Nano 33 BLE, Nano 33 BLE Sense) which use mbed framework and weren't compiling due to the code referencing a non-existing array.


runger1101001:
Tested on Nano 33 BLE and Feather Bluefruit nRF52832 board

runger1101001:
It changes only the nrf52 hardware support, I have tested it so I think it can be merged.

PR #141, opened 2021-11-27
define sizes and explicit values for all enums

runger1101001:
This commit adds defined sizes (uint8_t or int8_t in one case) and explicit values to all our enums.

This is due to the work on I2CCommander, and the need to do serialisation/deserialisation of these values in binary format.

While in theory the C++ compiler would probably have generated stable values, making them explicit guarantees this, and is also a form of documentation (e.g. makes it easy to see which enum constant has byte value 0x04 when looking at a Wireshark capture).

in addition there is a subtlety that enums are usually ints e.g. (sizeof(enumtype)==4 on 32 bit MCUs) but if you don't specify it isn't guaranteed. So when serialising, you could wind up with different sizes for the enum types on different architectures (or maybe due to compiler settings). Also that makes them subject to endianness problems (ugh :-( ). 
Luckily all our enums have <256 elements, so I have defined their size to 8bit integers which gets rid of all these problems.

### DIFF
--- a/src/common/base_classes/FOCMotor.h
+++ b/src/common/base_classes/FOCMotor.h
@@ -24,31 +24,31 @@
 /**
  *  Motiron control type
  */
-enum MotionControlType{
-  torque,//!< Torque control
-  velocity,//!< Velocity motion control
-  angle,//!< Position/angle motion control
-  velocity_openloop,
-  angle_openloop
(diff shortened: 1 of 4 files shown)


runger1101001:
Tested compile under ArduinoIDE for Arduino UNO and Tiny M4.
Tested compile & run under STM32F4.

PR #144, opened 2021-12-11
enable SPI sensor for RP2040, its now supported

runger1101001:
remove the ifdefs protecting the SPI code on RP2040, since it now works (due to updates in the RP2040 framework, nothing we did).


runger1101001:
uncontroversial, I think.

PR #149, opened 2022-01-07
Modify PWM timer configuration according to esp32 SDK 2.0.2

samguns:
The compilation fails for "_configureTimerFrequency" in "esp32_mcu.cpp" with the latest esp32 SDK. It seems expressif made a lot of changes for the MCPWM timer settings. After reading the reference manual and some code analysis, I came up with a patch for this problem.


askuric:
Thanks for your time, this will help us go a lot faster!
I'm merging!

PR #160, opened 2022-03-09
Fix example. M_PI does not exist, use _PI instead

gagarcr:
I think there is a typo in this example. M_PI is not defined, whereas _PI is defined in foc_utils.h


runger1101001:
Thank you for noticing this!

I think we can accept this pull request (but to the dev branch)

PR #161, opened 2022-03-09
Remove warnigs

gagarcr:
I am working in a port to another platform where the compiler keeps complaining about "unused paramerer warning". That is, when a function does not use all its arguments.

The fix is very simple and has no other implications. 

Also another warning when doing: 
char* separator = " "

That a simple cast removes it.


askuric:
Hey @gagarcr,
Thanks for the PR>
I am in, I've  used the macro `_UNUSED` for declaring the unused vars in the code. 
It looks like I've missed quiet a few. :D

Could you use the same macro instead of `(void)` ?
Also, the master is a sort of LTS version so we only accept the PRs for the dev branch.

gagarcr:
Sure, I havent noticed the _UNUSED macro. Will change as soon as possible. 

For future PRs I will use the dev branch.

PR #164, opened 2022-04-01
feat: add linear hall position sensing

nanoparticle:
Added example sketch to find linear hall ADC center offsets, plus tweak to BLDCMotor for continuous sensor updating. Actual sensor driver will go in the SimpleFOC drivers repo, this is just supporting infrastructure.


runger1101001:
My feeling is that we can add this, no problem. 

Calling sensor->update() regularly during the alignment is something we should do anyway!

runger1101001:
Another user who would like this same thing: https://github.com/simplefoc/Arduino-FOC/issues/173

runger1101001 (approved):
Looks good!

PR #165, opened 2022-04-02
STM32 driver - fixing software 6PWM

runger1101001:
Software 6-PWM looking good now.

Tested on STM32F401


runger1101001:
I will merge it, as it fixes bugs in what is already merged to dev.

PR #175, opened 2022-04-28
Added delay in serial print for stability

vbratasiuk:


runger1101001:
Seems like a good idea to me...

askuric:
LGTM :)

At least for this simple example.

PR #194, opened 2022-06-19
Correct the commands label to "target velocity"

yuyangwang-git:
I think there is a wrong commands label in this example, line80. 

According the doTarget function, line 32, this command change target_velocity, not the voltage.


askuric:
Hey @yuyangwang-git, thanks!
Ok for me, @runger1101001 you're ok with this?

runger1101001:
For sure!

PR #195, opened 2022-06-19
change getVelocity() to remember velocity

runger1101001:
Change in the way getVelocity() works.
It now "remembers" the last value calculated, and re-calculates it only when the delta-T is large enough (controllable through a new member field, default value is 100µs
 `float minDeltaT = 0.000100; // default is 100 microseconds, or 10kHz`

This will, I think, more elegantly handle the edge cases:
- getVelocity() called too often
- microseconds wrap

It does however introduce a kind of low-pass filter based on minDeltaT - one that was implicitly present in some form anyway, but is now explicit.

Please take a look and see if you agree with this change @askuric - we discussed it, but this might not be the way you imagined it?

### DIFF
--- a/src/common/base_classes/Sensor.cpp
+++ b/src/common/base_classes/Sensor.cpp
@@ -2,7 +2,7 @@
 #include "../foc_utils.h"
 #include "../time_utils.h"
 
-// TODO add an init method to make the startup smoother by initializing internal variables to current values rather than 0
+
 
 void Sensor::update() {
     float val = getSensorAngle();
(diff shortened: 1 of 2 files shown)


runger1101001:
Note: tested with gimbal motors on STM32G4 and SAMD21

runger1101001:
Oh, and also important to mention: the change also implicitly, through its deltaT check, computes the velocity at most once per call to update() - since the timestamp is only updated during update().
So getVelocity() now has the same behaviour as getAngle(), where the value remains stable within one loop iteration, i.e. requiring a call to update() before the value changes.

askuric:
I love it, its so simple and effective. That is more or less the solution I had in mind also. The only thing is that we might define this `minDeltaT` differently for different architectures. Like teensy/esp32/stm32g4/stm32f4 maybe go to 10us and the rest to 100us. 
But exposing this variable is probably the best way to resolve this, leave it to the user to find the best value. 

Im in for merging this :D
time_constriant

runger1101001:
Ok then lets do it!

PR #197, opened 2022-06-24
Add two linear halls sensor module.

XieMaster:
Hello!

I implemented adding two Linear Halls to run the motor.

Get raw analog value A&B --> filter --> map --> 
float angle = atan2(hall_filtered_a, hall_filtered_b) * 180 / PI;

Thank you!


askuric:
Hey @XieMaster,
Thanks for the contribution, we will merge this directly I think it a great contribution. However, I'm cannot test it as I do not have the hardware necessary so if someone is interested in testing it I would appreciate the feedback.

Also, I am not sure that I'll have enough time to document the new sensor for the release v2.2.3 so we might wait for the v2.2.4 to add it to the library.

dekutree64:
I have tested this, and in its current form it doesn't appear to work with anything but 1 pole pair. But if you paste in the electrical revolution tracking from nanoparticle's version, then it does work with my 12N14P motor. The range scaling can be useful to correct for unequal amplitude of sensor signals if they're not placed perfectly, but I think for speed and code size I prefer nanoparticle's version simplefoc/Arduino-FOC-drivers#12 which only subtracts the center value. Better to correct any amplitude difference by physical placement of the sensors.

PR #202, opened 2022-07-13
Portenta H7 6pin PWM pins?

Juanduino:
Hi
Like some of you have noticed. I have been doing a hardware development, trying to integrate the Portenta into a compact eMTB controller. https://github.com/Juanduino/MTBeDRIVE

My plan is to implement 6pin pwm. @runger1101001 How far did you get with it. Did you use the 80pin connectors or just the headers on the Portenta. I ques they must be on same timer in pairs?

Best regards

Juan-Antonio Søren

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -57,7 +57,7 @@ jobs:
 
           - arduino-boards-fqbn: STMicroelectronics:stm32:GenF1:pnum=BLUEPILL_F103C8 # bluepill - hs examples
             platform-url:  https://github.com/stm32duino/BoardManagerFiles/raw/main/package_stmicroelectronics_index.json 
-            sketch-names: bluepill_position_control.ino, stm32_i2c_dual_bus_example.ino, stm32_spi_alt_example.ino, stm32_current_control_low_side.ino
+            sketch-names: bluepill_position_control.ino, stm32_i2c_dual_bus_example.ino, stm32_spi_alt_example.ino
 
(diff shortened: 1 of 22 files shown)


Juanduino:
I see, according to this (https://www.st.com/resource/en/datasheet/stm32h747ag.pdf) the setting up of timers / advanced timers is the same for STM32G and STM32H.

Juanduino:
https://community.simplefoc.com/t/portenta-timer-for-6pwm/2268

runger1101001:
Hey, I'm all for adding the 6-PWM support for the portenta. I will get to it eventually, but I have so many things on my list...

So if someone were to provide a good implementation, I am sure we would accept it ;-)

I agree with you that the Portenta timers should be able to be initialised in the same way as a G4. The problem may be that Arduino framework for Portenta builds on mbed, so while our other PWM drivers for STM32 share the same code-base, the portenta driver has its own mbed-based implementation.
If the Arduino framework for Portenta includes HAL underneath mbed, and we can initialise the timers that way (bypassing mbed), then I think I would be in favour of unifying the STM32 PWM driver and just using HAL. 
If mbed is required, then the HAL code won't be helpful and new mbed based 6-PWM code will have to be written. Given the complexity of the H7's timers, this is a non-trivial task.

Even with HAL support it can be non-trivial. The H7 can assign its peripherals to either its M4 or M7 cores - so this is something that would have to be taken into consideration - is SimpleFOC running on the M4 core or the M7 core, and which timers are actually available?

runger1101001:
In the meantime, we should probably close this pull request... it is against the master branch (we only accept pull requests against the dev branch), and it will probably be some time before there is actual code to merge?

PR #203, opened 2022-07-25
typing error

lukaspfitscher:


runger1101001:
This pull request seems ok, but we only accept pull requests against the dev branch, not the master...

When I change the base branch to dev, it has a conflict. So I can't easily accept this as it is...

If its ok with you I'll just fix that typo with my next push to the dev branch. But if you'd like to contribute it yourself, please provide another pull request against the dev branch...

PR #210, opened 2022-08-24
B g431 b current sense fixes

sDessens:
Only tested on B-G431B with inline current sense.

### DIFF
--- a/src/current_sense/hardware_specific/esp32/esp32_mcu.cpp
+++ b/src/current_sense/hardware_specific/esp32/esp32_mcu.cpp
@@ -68,8 +68,6 @@ int adc_read_index[2]={0};
 
 // function reading an ADC value and returning the read voltage
 float _readADCVoltageLowSide(const int pin, const void* cs_params){
-  uint32_t raw_adc;
-
   mcpwm_unit_t unit = ((ESP32MCPWMCurrentSenseParams*)cs_params)->mcpwm_unit;
   int buffer_index = ((ESP32MCPWMCurrentSenseParams*)cs_params)->buffer_index;
   float adc_voltage_conv = ((ESP32MCPWMCurrentSenseParams*)cs_params)->adc_voltage_conv;
--- a/src/current_sense/hardware_specific/stm32/b_g431/b_g431_hal.cpp
+++ b/src/current_sense/hardware_specific/stm32/b_g431/b_g431_hal.cpp
@@ -1,10 +1,13 @@
 #include "../../../hardware_api.h"
 #if defined(ARDUINO_B_G431B_ESC1)
 
+#include "communication/SimpleFOCDebug.h"
+
 #include "stm32g4xx_hal.h"
 #include "stm32g4xx_ll_pwr.h"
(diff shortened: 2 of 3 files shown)


askuric:
Did you try using `LowsideCurretSense` instead of `InlineCurrentSense`? 
Because all the additions that you've made are already present in the `_driverSyncLowSide` function which is not called in for in-line class. I have moved them there in the last release to remove bg431 specific code from the driver code and because bg431 only has the low-side current sensing. So I would prefer not to use the inline code with it.

https://github.com/simplefoc/Arduino-FOC/blob/84a4aba4e83c50b2a3ec5c8eeb2cd763934f43da/src/current_sense/hardware_specific/stm32/b_g431/b_g431_mcu.cpp#L137

sDessens:
That's a good point. I feel I messed up with the original implementation, we called it inline current sense, but it really implements low side current sense. I'm confused why [the wiki](https://docs.simplefoc.com/inline_current_sense) claims inline current sensing support for this board, it seems that placing the required resistors would require very advanced soldering...

I tried using low side current sense, but it doesn't work. There are multiple problems with initialization and syncing, I will attempt a fix.

askuric:
Hey @sDessens,
I'll order the board and try to fix it as well. If you don't have time dont worry about it. This PR is already a great step in finding the real reasons for why it does not work.

The docs are wrong you are right!
I've written in that way because at some point we only had your implementation of the BG431 board which was done using the inline current sese (the only one supported at the time). 

So I've gone quickly through the code and can it be that the line:
```
LL_TIM_SetTriggerOutput(HardwareTimer_Handle[index]->handle.Instance, LL_TIM_TRGO_UPDATE);
```

has to be called before the  timer init?

I did not have any problems with this so far for f1, f4 and g4 low-side current sensing. They all work with the TRGO UPDATE setting 
https://github.com/simplefoc/Arduino-FOC/blob/1acfda1d22aac484ab231fafd729f480603fd854/src/current_sense/hardware_specific/stm32/b_g431/b_g431_mcu.cpp#L151

and repetition counter downsampling 
https://github.com/simplefoc/Arduino-FOC/blob/1acfda1d22aac484ab231fafd729f480603fd854/src/current_sense/hardware_specific/stm32/b_g431/b_g431_mcu.cpp#L144 

that are done after the timer init. They are called when the timer is stopped and once they are called the timer is resumed.

However, the bg431 code is the only one which for now uses the DMA so this might be the issue.

sDessens:
I managed to fix the problem, it turns out there was a null-pointer and this caused unpredictable behavior.

With this rebase, LowsideCurrentSense works but InlineCurrentSense does *not* work.

askuric:
Awesome, thanks!

This was my fault, I've refactored the code but as I did not have the board I did not test it properly!
Thanks a lot for taking the time to debug.

sDessens:
I feel the community might be slightly confused that we switched from inline to lowside current sensing on this board. Would it be a good idea to display a warning when the user initializes inlineCurrentSense on this board. For example by displaying a debug message in  _configureADCInline?

askuric:
Merging, thanks again!

PR #211, opened 2022-08-27
add routines for 1-pwm in driver code

runger1101001:
Add initial support (api and generic implementation only so far) for 1-pwm in the driver code.

1-PWM is not useful for BLDC or stepper control, but has other applications connected to our library:

- DC Motor control (stay tuned for new stuff in the drivers library)
- Active and regenerative braking
- PFC (power factor correction)
- and of course it can be used to dim the headlights of your robot :-D


askuric:
You're right, this is a good extension.

runger1101001:
:-) I'll merge it... I'll try to add some hardware driver implementations in the weeks to come, but first I will test the power-shield

PR #213, opened 2022-09-10
Set phase state (stm32)

sDessens:
I need this functionality for backemf sensing, since it's a nice isolated change I'm putting it up for review now.

Tested on the B-G431B, this is the only board I have. I'm not sure if this has the desired effect on boards with software 6pwm.

I'm not very happy with the API design, it relies on the user calling `_setPWM` shortly after `_setPhaseState`. Maybe remove `BLDCDriver::setPhaseState(sa, sb, sc)` and modify the signature of `setPwm(Ua, Ub, Uc)` to include the phase state, for example `setPwm(float Ua, float Ub, float Uc, int stateA, int stateB, int stateC)`. 9 times out of 10 you'll want to atomically change the pwm and phase state anyway. Maybe I'm overthinking things.

For testing, use `Trapezoid_120`. I believe `Trapezoid_150` is broken but that's out of scope for this pull request.

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -522,7 +522,7 @@ void BLDCMotor::setPhaseVoltage(float Uq, float Ud, float angle_el) {
         Ub -= Umin;
         Uc -= Umin;
       }
-
+      driver->setPhaseState(_ACTIVE, _ACTIVE, _ACTIVE);
       break;
 
     case FOCModulationType::SpaceVectorPWM :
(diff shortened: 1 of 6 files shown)


askuric:
Hey @sDessens,

Thanks for this. The PR has been merged to dev branch. I've readapted it to the new low-side api proposed by @runger1101001 in the #231. 

I've tested it and it seems to work well. Ill look in the trap 150, thanks for the heads up.

PR #221, opened 2022-10-10
Bg431b ADC extension

Igitigit2:
Now, that all the ADC stuff for the B-G431B-ESC1 board is in hardware specific files anyway, I see no value in not supporting the additional ADC channels for the bus voltage, temp sensor and poti. The code for these was proposed long time ago already by @HoeckDK on the SimpleFOC community forum. To read e.g. the temp sensor of the board, apply this PR and the call 

float Temp = _readADCVoltageInline(A_TEMPERATURE, currentSense.params);

It would be cool if this PR was accepted, this would safe me quite some time to apply it to all new versions over and over again...


Igitigit2:
Will create new PR based on dev branch...

PR #222, opened 2022-10-10
Added changes to use additional ADC channels

Igitigit2:
Now, that all the ADC stuff for the B-G431B-ESC1 board is in hardware specific files anyway, I see no value in not supporting the additional ADC channels for the bus voltage, temp sensor and poti. The code for these was proposed long time ago already by @HoeckDK on the SimpleFOC community forum. To read e.g. the temp sensor of the board, apply this PR and the call

float Temp = _readADCVoltageInline(A_TEMPERATURE, currentSense.params);

It would be cool if this PR was accepted, this would safe me quite some time to apply it to all new versions over and over again...

### DIFF
--- a/src/current_sense/hardware_specific/stm32/b_g431/b_g431_hal.cpp
+++ b/src/current_sense/hardware_specific/stm32/b_g431/b_g431_hal.cpp
@@ -95,7 +95,7 @@ void  MX_ADC1_Init(ADC_HandleTypeDef* hadc1)
   hadc1->Init.EOCSelection = ADC_EOC_SINGLE_CONV;
   hadc1->Init.LowPowerAutoWait = DISABLE;
   hadc1->Init.ContinuousConvMode = DISABLE;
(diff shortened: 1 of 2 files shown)


askuric:
I think this is perfectly reasonable. It is not really in the spirit of the simplefoc api but at least it will enable using this board a bit more optimally in the mean time. Did you notice any performance issues when you do 5 conversions instead of 3? 
@runger1101001 do you have any thoughts about this?

Igitigit2:
No, I did not notice any performance issues, but haven‘t tried it a lot with the latest dev branch. 
@runger1101001, of course, the idea is to have it specific to the bg431 board. Since all the ADC stuff for this board is in board specific files, the PR should not make SimpleFOC in general less portable/ compatible.

askuric:
OK, I'm merging this for v.2.2.3

Igitigit2:
Excellent, thank you!

PR #224, opened 2022-10-23
Method configure4PWM of esp8266_mcu.cpp bug fix

18301328831:


bwees:
Can confirm this is a needed fix. Please merge!

runger1101001:
I've changed the base to the "dev" branch, as we only merge PRs there.

Looks like this PR picked up another change to Sensor.h as well, which for some reason was made against the master branch.

Nonetheless, I think this can be merged.

Thank you very much for it!

PR #230, opened 2022-11-20
STM32 driver improvements & PWM sensor bugfix

runger1101001:
- add 1-PWM mode to STM32 driver
- make linkDriver() function virtual in FOCMotor so we can override it
- add polarity control via -D SIMPLEFOC_PWM_ACTIVE_HIGH=true/false to STM32 driver
- use the min_raw_count in PWM sensor


askuric:
It looks good to me

PR #251, opened 2023-02-13
compile time fix for rp2040 - include hardware/pwm.h

mschuldt:
Without this include I get many declaration errors such as:
`error: 'pwm_gpio_to_slice_num' was not declared in this scope`


runger1101001:
Hey, this change is already present on the dev branch of the library, and will be part of the next release.

But thank you very much for reporting it!

PR #253, opened 2023-02-20
add constructor taking mV per Amp to current sense

runger1101001:
as per Valentine's request. It makes sense for hall based current sensors, which don't work with a shunt resistor.


runger1101001:
All tests now pass, so I'll merge this over

PR #261, opened 2023-03-21
dev´s

Juanduino:


Juanduino:
Feel free to integrate it, if you have some ideas?

Juanduino:
This is compiling on my end

runger1101001:
Oooof....

In terms of the CORDIC stuff, this won't work like this. I don't want to merge anything related to CORDIC into the main library. It's not useful to 99.8% of our users, and it will make all the test-builds fail if left in like this.
To get the CORDIC functionality, I have recently merged to the dev branch changes which enable you to supply your own versions of the _sin() and _cos() functions.

In terms of the 8-PWM stuff, I can't follow all these changes :-)
It should be possible to do it, but the model is not clear to me:
- do you expect 8 PWM channels on the same timer? I don't think the MCUs can do it?
- do you expect to use complimentary channels (CHx and CHxN)?
- or does none of this matter, and you just want 4 pairs of pins, with each pair on the same timer?
- I see you expect dead-time insertion? I think its needed for 8-PWM?

These are relevant questions because the code can't just work on any PWM pins - I believe in the 8-PWM configuration requires complementary PWMs, like in 6-PWM? That means the complimentary signals *have* to come from the same timer. But they don't necessarily have to use CHx CHxN pins - it can also work with CHx CHy - but then there is no HW dead-time, only software. Also you can't mix and match hardware dead-time and software dead-time - it's either one or the other for all the pins. And you can't use CHxN on the high side, e.g. you can only use those when doing all HW dead-time.
So there are loads of complexities and constraints :-(

askuric:
Hey @Juanduino, this starts to be a bit old. I'm closing it for now.

PR #267, opened 2023-03-31
inital implementation of the velocity feed forward

askuric:
- An initial implementation of the velocity feed forward in order to allow for more advanced planning strategies 

For example for trapezoidal velocity profile: 
https://community.simplefoc.com/t/3d-printed-robot-arm-acceleration-deceleration-planner-algorithm/958/39

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -418,7 +418,8 @@ void BLDCMotor::move(float new_target) {
       // angle set point
       shaft_angle_sp = target;
       // calculate velocity set point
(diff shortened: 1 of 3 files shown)


askuric:
Hey @dekutree64,
You're absolutely right!
I'm so glad that it stayed in this PR, thanks for taking the time to go in such detail. I'll update it right away!

askuric:
I am not sure how, but this is already inside the dev branch. :D
I must have done some accidental pushes. 
So, there is no point in leaving this one open any more, I'm merging it.

Issue #270, opened 2023-05-01
Refactoring interrupt-based sensor code


dekutree64:
When working on my experimental [SmoothingSensor](https://community.simplefoc.com/t/smoothingsensor-experimental-sensor-angle-extrapoltion/3105) class, I became aware that many of the SimpleFOC sensor classes can give sporadic bad readings due to interrupts occurring in the middle of operations. And to make matters worse, the Arduino library does not provide any way to temporarily disable interrupts and then restore to the previous state (noInterrupts() does not return the previous state).

Note: I’m simultaneously working on getting more accurate timestamps for angle prediction by SmoothingSensor.

My proposed solution for HallSensor is to do away with all the specialized get functions except for getSensorAngle, so everything accesses the Sensor base class state variables except during update(), which is only ever called at a time when interrupts are enabled so I can call the disable/enable functions without worrying about the previous state. One option would be to make a specialized update() that blocks interrupts, like this:
```
void HallSensor::update() {
  noInterrupts();
  Sensor::update();
  angle_prev_ts = pulse_timestamp;
  interrupts();
}
```
So the call to HallSensor::getSensorAngle inside Sensor::update is protected from interrupts modifying electric_sector after electric_rotations is loaded from memory, and the timestamp is also guaranteed to be the correct one for the angle reading.

But due to the fact that two other sensors (MagneticSensorPWM and Encoder) need similar treatment, I think it would be better to move pulse_timestamp to the base class, and do like this:
```
void Sensor::update() {
    noInterrupts();
    float val = getSensorAngle();
    angle_prev_ts = _isset(pulse_timestamp) ? pulse_timestamp : _micros();
    interrupts();
    float d_angle = val - angle_prev;
    // if overflow happened track it as full rotation
    if(abs(d_angle) > (0.8f*_2PI) ) full_rotations += ( d_angle > 0 ) ? -1 : 1; 
    angle_prev = val;
}
```
So if the subclass uses pulse_timestamp, it is transferred to angle_prev_ts at the same time as the other state variables are updated, else _micros() is called as before.

But it’s a little iffy because it means you can’t use interrupt-based communication methods inside getSensorAngle. I don’t think the Arduino I2C and SPI libraries use interrupts, but it’s still something that could potentially cause trouble in the future, so I would add a note about it in the comment description of getSensorAngle.

With this approach, MagneticSensorPWM would only need this small change to handlePWM:
```
    if (!digitalRead(pinPWM)) {
[...]

PR #274, opened 2023-05-22
Fix for issue #270

dekutree64:
Overhaul of interrupt-based sensors to eliminate bad readings due to interrupts modifying variables in the middle of get functions.

All sensor classes are now required to use the base class state variables, and do their angle reading once per frame in the update() function rather than using the virtual get functions to return real-time updated readings. I have chosen to make an exception for getVelocity() and allow it to access volatile variables as well, although it may be better to merge it into update() instead. Interrupts should be disabled/enabled as necessary in the update() and getVelocity() functions, and preferably nowhere else. The Arduino library provides no mechanism to restore the previous state of interrupts, so use of the unconditional enable should be kept to as few locations as possible so we can be reasonably sure that it won't be called at a time when interrupts should remain disabled.

Additional changes to specific sensors:

HallSensor: Removed the max_velocity variable because it was a quick fix for bad velocity readings that were coming from this interrupt problem, which should no longer occur. Also changed the condition for "velocity too old" from (_micros() - pulse_timestamp) > pulse_diff to (_micros() - pulse_timestamp) > pulse_diff*2, because any deceleration would cause inappropriate reporting of zero velocity.

MagneticSensorPWM: Unrelated to the interrupt problem, timestamp is now saved from the rising edge of the PWM pulse because that's when the angle was sampled, and communicating it takes a significant and variable amount of time. This gives more accurate velocity calculations, and will allow extrapolating accurate angles using the new SmoothingSensor class.

### DIFF
--- a/src/common/base_classes/Sensor.h
+++ b/src/common/base_classes/Sensor.h
@@ -42,6 +42,7 @@ enum Pullup : uint8_t {
  * 
  */
 class Sensor{
+	friend class SmoothingSensor;
     public:
         /**
          * Get mechanical shaft angle in the range 0 to 2PI. This value will be as precise as possible with
--- a/src/sensors/Encoder.cpp
+++ b/src/sensors/Encoder.cpp
@@ -99,29 +99,21 @@ void Encoder::handleIndex() {
 }
 
 
+// Sensor update function. Safely copy volatile interrupt variables into Sensor base class state variables.
 void Encoder::update() {
-    // do nothing for Encoder
+  noInterrupts();
+  // TODO: numerical precision issue here if the pulse_counter overflows the angle will be lost
+  full_rotations = pulse_counter / (int)cpr;
+  angle_prev = _2PI * ((pulse_counter) % ((int)cpr)) / ((float)cpr);
+  angle_prev_ts = pulse_timestamp;
+  interrupts();
 }
 
 /*
 	Shaft angle calculation
 */
 float Encoder::getSensorAngle(){
-  return getAngle();
-}
-// TODO: numerical precision issue here if the pulse_counter overflows the angle will be lost
-float Encoder::getMechanicalAngle(){
-  return  _2PI * ((pulse_counter) % ((int)cpr)) / ((float)cpr);
-}
-
-float Encoder::getAngle(){
-  return  _2PI * (pulse_counter) / ((float)cpr);
-}
-double Encoder::getPreciseAngle(){
-  return  _2PI * (pulse_counter) / ((double)cpr);
-}
-int32_t Encoder::getFullRotations(){
-  return  pulse_counter / (int)cpr;
+  return _2PI * (pulse_counter) / ((float)cpr);
 }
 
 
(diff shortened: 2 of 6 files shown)


askuric:
Hey @dekutree64,

Thanks for this, I can see the benefits from this addition in terms of velocity smoothness for sure. 
I had no idea that `noInterrupts` exists in Arduino IDE .

Now, I am not sure that the `noInterrupts` and `interrupts` is the best approach when it comes to Encoder. Hall sensor is fine because it is absolute and even if we miss some interrupts at the first next one after re-enabling the interrupts we will no the exact position of the motor. 

However the encoder is not like that, and disabling interrupts will make it lose counts (provided high enough velocity) . I am not sure what would be best solution here be though. Maybe disable the interrupts only for fetching the necessary volatile variables `pulse_counter` and `pulse_timestamp` and then enabling them for the rest of the velocity calculation. 

I am not sure, but I am a bit worried that it could introduce issues for encoders which are arguably the most used sensors with the simplefoc. What are your experiences, what kind of benefits have you seen for encoders of disabling the interrupts? 
Smoother velocity, less jitters?

runger1101001:
> The Arduino library provides no mechanism to restore the previous state of interrupts, so use of the unconditional enable should be kept to as few locations as possible so we can be reasonably sure that it won't be called at a time when interrupts should remain disabled.

I think we have to think about this one carefully - can it even happen?
Also what exactly happens - is the interrupt merely delayed, or is it skipped? I need to look into this, I guess.
The reason is that interrupts are used for example also in the current sensing in some cases, or by the I2CCommander, and of course they might be used in other user code.

dekutree64:
The re-enabling problem would mainly show up if we put disable/enable blocks in getAngle() and similar functions. Then if user code ever called them from inside an interrupt handler, it would allow nested interrupts which would likely crash the system. Or if they were called during an interrupt blocking section of another function, it would re-enable interrupts sooner than intended. That's less bad, but still can cause loss of motor control in some cases like the HallSensor::getVelocity() divide by zero we patched previously.

Or if you were referring to the missed encoder pulses, then yes, a single blocked interrupt will only be delayed, not missed. Each interrupt source has a bit in a register that gets set when it triggers, and will interrupt the CPU if interrupts are enabled. You can have multiple interrupts from different sources pending, and they will fire off in a pre-determined priority each time interrupts are re-enabled. But if the same interrupt triggers again while its bit is already set, then by the time it gets handled there's no way to tell that it had been set more than once. That's when you get missed encoder pulses, which would only be at very high speed and resolution, but still conceivable.

And it's good practice to minimize interrupt blocking time in general, so I went ahead and did it for HallSensor too. It also allowed reformatting getVelocity() back closer to how it was originally (not that it makes any functional difference).

P.S. Was it a mistake to close this pull request and open a new one after making the changes? This is my first time using GitHub, and in hindsight there was probably a way to edit this one to keep the conversation all in one place...

runger1101001:
No its fine but I've moved the conversation to the active PR... these things happen because we were both commenting and doing stuff at the same time :-)

PR #276, opened 2023-05-22
Fix for issue #270

dekutree64:
Overhaul of interrupt-based sensors to eliminate bad readings due to interrupts modifying variables in the middle of get functions.

All sensor classes are now required to use the base class state variables, and do their angle reading once per frame in the update() function rather than using the virtual get functions to return real-time updated readings. I have chosen to make an exception for getVelocity() and allow it to access volatile variables as well, although it may be better to merge it into update() instead. Interrupts should be disabled/enabled as necessary in the update() and getVelocity() functions, and preferably nowhere else. The Arduino library provides no mechanism to restore the previous state of interrupts, so use of the unconditional enable should be kept to as few locations as possible so we can be reasonably sure that it won't be called at a time when interrupts should remain disabled.

Additional changes to specific sensors:

HallSensor: Removed the max_velocity variable because it was a quick fix for bad velocity readings that were coming from this interrupt problem, which should no longer occur. Also changed the condition for "velocity too old" from (_micros() - pulse_timestamp) > pulse_diff to (_micros() - pulse_timestamp) > pulse_diff*2, because any deceleration would cause inappropriate reporting of zero velocity.

MagneticSensorPWM: Unrelated to the interrupt problem, timestamp is now saved from the rising edge of the PWM pulse because that's when the angle was sampled, and communicating it takes a significant and variable amount of time. This gives more accurate velocity calculations, and will allow extrapolating accurate angles using the new SmoothingSensor class.

Sensor: Added friend class declaration for SmoothingSensor. Alternatively we could make prev_angle_ts public, and I could use getMechanicalAngle() and getFullRotations() to access the other state variables.

### DIFF
--- a/src/common/base_classes/Sensor.h
+++ b/src/common/base_classes/Sensor.h
@@ -42,6 +42,7 @@ enum Pullup : uint8_t {
  * 
  */
 class Sensor{
+	friend class SmoothingSensor;
     public:
         /**
          * Get mechanical shaft angle in the range 0 to 2PI. This value will be as precise as possible with
--- a/src/sensors/Encoder.cpp
+++ b/src/sensors/Encoder.cpp
@@ -101,12 +101,14 @@ void Encoder::handleIndex() {
 
 // Sensor update function. Safely copy volatile interrupt variables into Sensor base class state variables.
 void Encoder::update() {
+  // Copy volatile variables in minimal-duration blocking section to ensure no interrupts are missed
   noInterrupts();
-  // TODO: numerical precision issue here if the pulse_counter overflows the angle will be lost
-  full_rotations = pulse_counter / (int)cpr;
-  angle_prev = _2PI * ((pulse_counter) % ((int)cpr)) / ((float)cpr);
   angle_prev_ts = pulse_timestamp;
+  long copy_pulse_counter = pulse_counter;
   interrupts();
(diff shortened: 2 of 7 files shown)


runger1101001:
> The Arduino library provides no mechanism to restore the previous state of interrupts, so use of the unconditional enable should be kept to as few locations as possible so we can be reasonably sure that it won't be called at a time when interrupts should remain disabled.

I think we have to think about this one carefully - can it even happen?

Also what exactly happens - is the interrupt merely delayed, or is it skipped? I need to look into this, I guess.
The reason is that interrupts are used for example also in the current sensing in some cases, or by the I2CCommander, and of course they might be used in other user code.

runger1101001:
Looking into it a little bit I would say interrupts are normally delayed rather than just cancelled.

To check the state of the interrupt flag, there could be options outside of Arduino framework.

Thinking about whether it is needed:
- Arduino model is single threaded, so in regular code  interrupts cannot be disabled unless the user has done so in the call-chain. So normally, user should be aware he's already disabled interrupts. I think its probably not something we need to consider.

askuric:
Hey guys, 

I am OK with these changes, the influence to the execution should be minimal.  However, we'll need to test thoroughly. :D

Thanks again @dekutree64!

runger1101001:
Then lets merge it :-)

PR #280, opened 2023-06-18
Bug-Fixes, including BLDCMotor KV init bug

runger1101001:
RP2040: move default sample rate to header
Move default sample rate of 20kHz to header file so it can be changed by users...

BLDCMotor - KV was being initialised even when _KV was set to NOT_SET


runger1101001:
I think these can be merged...

PR #285, opened 2023-06-29
change sine implementation to deku65i version

runger1101001:
We have already discussed it - this is now the PR for the sine optimisation.

It includes the following changes:

1)
- introduce a new _sincos() function which computes sin and cos at the same time
    - based on this significant further optimisation is possible for users that want to implement it
    - for example, STM32 CORDIC can return both sin and cos in about 25 clock cycles...
- default implementation just calls _sin() and _cos() internally
- code that used both sine and cosine of the same value has been updated to use the new function

2)
- default implementation of the _sin() function has been changed to the "deku65i" version
    - this uses a 65 entry LUT of 16bit values, and 8bits of fractional interpolation between neighbouring values
    - it uses the full precision of the 16bit values (rather than 10000x)
    - it can operate on any input angle (no need to normalize)
    - it is faster than the original version (by about 20% for just sine, or about 4.5x faster if you include normalizeAngle)
    - it is more precise than the original version (RMS precision compared to stdlib for new version 0.0000648 compared to 0.00161161 for original. That's in the range -PI,PI for 3217 steps of 1/1024 in size)
- remove normalizeAngle before calling sin/cos, since it isn't needed any more

Many details can be found in this thread in the forums, if one has the patience to follow it: https://community.simplefoc.com/t/embedded-world-2023-stm32-cordic-co-processor/3107 

Tested on SAMD21, where it seems to give 5-10% overall performance boost to my main loop (velocity mode, sine-commutation, no current sense).

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -539,12 +539,8 @@ void BLDCMotor::setPhaseVoltage(float Uq, float Ud, float angle_el) {
     case FOCModulationType::SinePWM :
       // Sinusoidal PWM modulation
       // Inverse Park + Clarke transformation
+      _sincos(angle_el, &_sa, &_ca);
 
-      // angle normalization in between 0 and 2pi
-      // only necessary if using _sin and _cos - approximation functions
-      angle_el = _normalizeAngle(angle_el);
-      _ca = _cos(angle_el);
-      _sa = _sin(angle_el);
       // Inverse park transform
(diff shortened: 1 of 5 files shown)


askuric:
Nice!
I'm in and I'm looking forward to testing it.

runger1101001:
Perfect, so I will merge it, and I'll also ask in the forum thread if people want to test the dev branch on their setups...

PR #286, opened 2023-07-01
Expose last wire.endTransmission code

padok:
Hey there!

I was wondering if we could somehow gain access to the error codes returned by the Wire library on endTransmission().

I noticed that when my AS5048B is somehow disconnected, the motor locks up at full voltage/torque. In my case, the motor got hot pretty quickly, and the 3D printer holder melted a bit. With the suggested code, I was now able to solve this by entering an error state:
```
  if (sensor.error)
  {
    Serial.println("Wire error");
    motor_error = true;
    motor.disable();
    driver.disable();
  }
```

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,9 @@ Please do not hesitate to leave an issue if you have problems/advices/suggestion
 
 Pull requests are welcome, but let's first discuss them in [community forum](https://community.simplefoc.com)!
 
-If you'd like to contribute to this porject but you are not very familiar with github, don't worry, let us know either by posting at the community forum , by posting a github issue or at our discord server.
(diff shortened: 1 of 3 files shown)


runger1101001:
I think this makes sense... if its ok with you I will rename the variable and change the comments a bit after merging it.

I'm changing the base to the dev branch. We only accept PRs against the dev branch...

padok:
Sorry for not using dev as a base, I just merged it, so it should be fine now!

Yes, feel free to edit anything you like!

I also have a suggestion for the "Github Issues & Pull requests" section of the README, and I fixed a typo in it.

runger1101001:
I've now merged it. Thanks a lot for the contribution and the idea!

PR #289, opened 2023-07-05
Support Arduino UNO R4 Minima and WiFi - PWM

runger1101001:
Adds support for the new Arduino UNO R4 boards based on Renesas MCUs.

This PR includes PWM support but not yet current sensing.

The MCUs will be nice for SimpleFOC:
- Several PWM pins, including those pins that do PWM on the original UNO
- 48MHz ARM Cortex M4 MCU with FPU
- PWM generation with either polarity
- hardware dead-time is possible in 6-PWM mode
- low side current sensing should be possible


runger1101001:
Note: this PR is not quite ready to merge yet. I still have to add support for 6-PWM and do some more testing.

runger1101001:
This is now working well enough in my tests that we can merge it and see if the UNO R4s work well with our sensors, examples, etc...

PR #290, opened 2023-07-05
add ESP32 output inversion support

mcells:
This PR adds support for switching driver polarity when using the MCPWM driver using the SIMPLEFOC_PWM_HIGH/LOWSIDE_ACTIVE_HIGH and SIMPLEFOC_PWM_ACTIVE_HIGH defines.

I have successfully used the 6 pwm inverted lowside modified code for my motor driver and just added the missing cases in.
The output of both 6 and 3 pwm modes looks correct to me. I didn´t try the stepper modes, but they should work the same. 

Are the proposed changes ok, or have I missed something?

### DIFF
--- a/src/drivers/hardware_specific/esp32/esp32_mcu.cpp
+++ b/src/drivers/hardware_specific/esp32/esp32_mcu.cpp
@@ -41,7 +41,7 @@ void _configureTimerFrequency(long pwm_frequency, mcpwm_dev_t* mcpwm_num,  mcpwm
 
   mcpwm_config_t pwm_config;
   pwm_config.counter_mode = MCPWM_UP_DOWN_COUNTER; // Up-down counter (triangle wave)
-  pwm_config.duty_mode = MCPWM_DUTY_MODE_0; // Active HIGH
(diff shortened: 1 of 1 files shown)


runger1101001:
This looks great to me, thank you very much!

Did you also test that the changes when using the standard setting still give the standard behaviour?

mcells:
Thanks!
Just tested it again, the default behavior is still everything active high. 

SIMPLEFOC_PWM_HIGH/LOWSIDE_ACTIVE_HIGH overrides SIMPLEFOC_PWM_ACTIVE_HIGH only when _configureTimerFrequency is called with the deadtime parameter set (currently only done by 6 pwm). This is the case, even when deadtime is set to 0.

If the defines in hardware_api.h are removed, PIO & Arduino won´t compile, instead of just treating them as false.

Perhaps it would be a good idea to check for these defines in all drivers and throw an error if a driver doesn´t implement setting them to nonstandard values?

runger1101001:
Thank you for checking!

> Perhaps it would be a good idea to check for these defines in all drivers and throw an error if a driver doesn´t implement setting them to nonstandard values?

Yes, this would be a very good idea! I will try to add it with the next few commits

runger1101001:
Thank you very much for your contribution! It's really appreciated. As you can imagine, adding features across all the different MCU architectures we support is a lot of work (making test setups each time!), so it is really a big help when generous people (like you) step in and help us out!

PR #299, opened 2023-08-22
Fixing errors with new Arduino framework versions

runger1101001:
Fixing some small compiler errors and special cases as caused by framework upgrades...


runger1101001:
This fixes #301 and some other problems.

PR #305, opened 2023-09-03
Add ESP32 Phasestate with sw-deadtime

mcells:
Adds Phasestate for the MCPWM driver.
I had to use software deadtime for this, so I added a define to (optionally) switch back to hardware-dt, should the need arise for some reason. 
The 6pwm functionality works nicely and the waveforms of the 3pwm output still seem to do what they should.

There is still one thing I am not quite sure about. IIrc, at least one of the existing sw-dt implementations turns the high/low phases completely on/off, when the requested dutycycle is 100 or 0 percent (instead of `100 - deadtime` or `0 + deadtime`). However I can´t find that anymore, so I left this out here too.

### DIFF
--- a/src/drivers/hardware_specific/esp32/esp32_driver_mcpwm.h
+++ b/src/drivers/hardware_specific/esp32/esp32_driver_mcpwm.h
@@ -81,6 +81,7 @@ typedef struct ESP32MCPWMDriverParams {
   mcpwm_unit_t mcpwm_unit;
   mcpwm_operator_t mcpwm_operator1;
   mcpwm_operator_t mcpwm_operator2;
+  float deadtime;
 } ESP32MCPWMDriverParams;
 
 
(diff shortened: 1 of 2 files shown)


runger1101001:
Thanks a lot for this. I am wondering if we should still include it in this release or the next one?

Regarding the turning on/off completely, you are right we are not consistent here. For example, I think the ATMega MCU drivers internally use analogWrite to set the PWM (which does this check for 0/100%) when not in 6-PWM mode, and they account for the 0/100% case explicitly in 6-PWM too.

runger1101001:
Ok, my feeling is that we could merge it in this release, but we set SIMPLEFOC_ESP32_HW_DEADTIME to true by default.
If people want to use the new feature, they can change the build flag...

That keeps the same behaviour as so far in this release, because my concern is that it would otherwise change the behaviour for all users, I'd like to test it a bit more first before we do that.

What do you think?

mcells:
That sounds like a good compromise. Although I am pretty certain that there are no bugs, it indeed seems a bit rushed to immediately release everything to main.

Regarding changed behaviour, the only difference is the added functionality and the 'issue' from above. 
And I just remembered that the hw-dt too turns off/on the phases completely, because it does deadtime insertion by delaying the switching.

runger1101001:
Thank you very much for this!! merging it now :-)

PR #309, opened 2023-09-13
SVPWM with midpoint clamp

Candas1:
Hi  @runger1101001 and @askuric,

Hope this is the right way to contribute.
This is what I proposed [here](https://community.simplefoc.com/t/embedded-world-2023-stm32-cordic-co-processor/3107/165?u=candas1).

Before the change: 
RAM:   [=         ]   9.2% (used 4536 bytes from 49152 bytes)
Flash: [===       ]  27.1% (used 71076 bytes from 262144 bytes)
loopfoc takes up to 220us with SVPWM 

After the change:
RAM:   [=         ]   9.2% (used 4536 bytes from 49152 bytes)
Flash: [===       ]  26.6% (used 69676 bytes from 262144 bytes)
loopfoc takes up to 156us with SVPWM
loopfoc takes up to 152us with SinePWM

I only saw a very small difference in magnitude between the old and new SVPWM without center modulation.
With center modulation it's exactly the same.

I decide to just add SVPWM to the SinePWM code and not duplicate all the code.


runger1101001:
I'm on-board with this. The question is whether we want to put it in this release (2.3.1) which is very close to happening, or merge it to the dev branch afterwards, for the 2.4.0 release?

Candas1:
I wouldn't want to mess up with the 2.3.1 release if there is not enough feedback about my change.
I will continue using the dev branch after the release.

runger1101001:
Just waiting for the Arduino library manager to pick up the new release, make sure everything is ok, and then I will merge this to the dev branch...

Candas1:
Take your time. Thanks for the new release.

PR #312, opened 2023-09-21
Update to correct 6pwm driver parameters.

endreSTT:


askuric:
Hey @endreSTT, 
Did you just add spaces here or are there some other modifications that I'm missing? 😄

endreSTT:
Looks like I only added some nice spaces yes. But I believed I changed this:
BLDCDriver6PWM driver = BLDCDriver6PWM(INH_A,INH_A, INH_B,INH_B, INH_C,INL_C, EN_GATE);


________________________________

runger1101001:
Hey, thanks for this one, but I think the INH_A,INH_A error was fixed in parallel, and the difference in spacing is a matter of opinion :-)

We very much appreciate every contribution, but in this case I think the essential part of the change is already in the code.

If its ok I will close this PR now.

PR #322, opened 2023-09-27
renesas: disable capture interrupt before open()

facchinm:
Fixes https://github.com/arduino/ArduinoCore-renesas/issues/139

Renesas fsp structures must be properly initialized entirely since `memset`ting all to 0 makes some values "valid" (in this case, the capture irq was firing as 0 instead than the agt interrupt, leading to a null pointer crash)


runger1101001:
Thank you so much for this! It makes complete sense.

What I don't understand is why only the WiFi was affected, an not the Minima.

I will merge to our dev branch and test it.

runger1101001:
Thanks @facchinm ❤️ this has indeed fixed the problem.

runger1101001:
https://github.com/arduino/ArduinoCore-renesas/issues/139 could be closed but I am still blocked from commenting on it / closing it

PR #326, opened 2023-10-23
Move Ualpha and Ubeta to FOCMotor

Candas1:
This is just about moving Ualpha and Ubeta to FOCMotor so it's available to the flux observer sensor.


Candas1:
I just forgot it's also used [here](https://github.com/simplefoc/Arduino-FOC/blob/f9e9a2d29e912957bd75302d4f9a83a1cdff37a5/src/StepperMotor.h#L76)
I keep forgetting about stepper motors.
It could be removed from StepperMotor.h then.

runger1101001:
Hmmm - it looks here like a bunch of the AVR checks are failing. I think we have to look into why before we can merge it...

Candas1:
> Hmmm - it looks here like a bunch of the AVR checks are failing. I think we have to look into why before we can merge it...

Welcome back and thanks for merging the PRs.
As I mentioned I forgot to update StepperMotor.h, but I am not sure if it's the reason compilation fails, I will investigate.

I am trying to keep each PR minimal, I hope it's ok.

Candas1:
I tried removing Ualpha and Ubeta from StepperMotor.h also and initializing it in FOCMotor.cpp, but no luck.
That looks like a very random issue.

I will investigate more.

runger1101001:
Weird, right? I haven't really figured out what it is...

But in the meantime I have found and fixed the various esp32 compile problems, of which there were several. None of them had anything to do with your changes, as expected.

runger1101001:
If you merge over my recent changes I think that will fix the ESP32 side of things, at least

Candas1:
Merges all the last dev changes fixed ESP32 compilation as you said, but avr is still failing.
So this issue is not so random as I thought.

runger1101001:
I think it is this: https://stackoverflow.com/questions/8188849/avr-linker-error-relocation-truncated-to-fit

askuric:
I think the avr issue is fixed so I'm merging this. :D

PR #327, opened 2023-10-23
STM32F1 Adc calibration

Candas1:
Adding ADC calibration for STM32F1.

**Before:**
RAM:   [=         ]   9.0% (used 4444 bytes from 49152 bytes)
Flash: [===       ]  26.6% (used 69712 bytes from 262144 bytes)

offset_ia = 1.574506
offset_ib = 1.566360

**After:**
RAM:   [=         ]   9.0% (used 4444 bytes from 49152 bytes)
Flash: [===       ]  26.8% (used 70208 bytes from 262144 bytes)

offset_ia = 1.630240
offset_ib = 1.623240

Slightly increases memory usage because of the use of an additional hal function, but should make all the adc measurements closer to reality.
The only risk I see are:
- a user has saved the phase current offsets without adc calibration in the code or in eeprom and would measure phase current with adc calibration, but I don't think it's common practice
- a user tweaked the gains to match an internal measurement, now the measurements won't match anymore


runger1101001:
I will merge it, the STM32 tests like it :-)

PR #329, opened 2023-10-23
Faster atan2

Candas1:
Shamelessly copied, even the credits (see the comments)

**Before:**
RAM:   [=         ]   9.0% (used 4444 bytes from 49152 bytes)
Flash: [===       ]  26.9% (used 70592 bytes from 262144 bytes)

up to 48us

**After:**
RAM:   [=         ]   9.0% (used 4444 bytes from 49152 bytes)
Flash: [===       ]  26.7% (used 69912 bytes from 262144 bytes)

up to 28us

Reduces memory usage and almost twice as fast on a STM32F1
It's not used in the SVPWM anymore but can be useful for other features.
I am using it in the flux observer sensor.

### DIFF
--- a/src/common/foc_utils.cpp
+++ b/src/common/foc_utils.cpp
@@ -44,6 +44,33 @@ __attribute__((weak)) void _sincos(float a, float* s, float* c){
   *c = _cos(a);
 }
 
+// fast_atan2 based on https://math.stackexchange.com/a/1105038/81278
+// Via Odrive project
+// https://github.com/odriverobotics/ODrive/blob/master/Firmware/MotorControl/utils.cpp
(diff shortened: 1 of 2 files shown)


runger1101001:
Hmmm...

```
/home/runner/Arduino/libraries/Arduino-FOC/src/current_sense/hardware_specific/esp32/esp32_adc_driver.cpp:9:10: fatal error: esp_intr.h: No such file or directory
    9 | #include "esp_intr.h"
      |          ^~~~~~~~~~~~
compilation terminated.
```

That doesn't really look like it has to do with your change...

So I will merge this, it is compiling for all other platforms except ESP32.

PR #339, opened 2023-11-12
Feature: FIX 2 phase measurements resulting in bad DQ currents

pglr:
inside the common/base_classes/CurrentSense.cpp there was an error with the if-else statement for both getCurrent methods. 
This leads to a very noisy DQ current if only two phase currents are measured. Either a&b or c&b lead to this issue:

image

after the FIX OR using a&c:

image


runger1101001:
I think you and Candas have found an important bug. Thanks very much for this!

Candas1:
For your information this also saves me 10us loopfoc duration on STM32F1 😍​

Issue #311, opened 2023-09-19
[FEATURE] Foc calculation refactoring


Candas1:
Hi,

This is not adding a feature, but could make further changes easier.
There is duplicate code [here](https://github.com/simplefoc/Arduino-FOC/blob/05954cb9691534dd478407505d810a68f1673a5e/src/common/base_classes/CurrentSense.cpp#L14) and [here](https://github.com/simplefoc/Arduino-FOC/blob/05954cb9691534dd478407505d810a68f1673a5e/src/common/base_classes/CurrentSense.cpp#L55)

``` // calculate clarke transform
    float i_alpha, i_beta;
    if(!current.c){
        // if only two measured currents
        i_alpha = current.a;  
        i_beta = _1_SQRT3 * current.a + _2_SQRT3 * current.b;
    }if(!current.a){
        // if only two measured currents
        float a = -current.c - current.b;
        i_alpha = a;  
        i_beta = _1_SQRT3 * a + _2_SQRT3 * current.b;
    }if(!current.b){
        // if only two measured currents
        float b = -current.a - current.c;
        i_alpha = current.a;  
        i_beta = _1_SQRT3 * current.a + _2_SQRT3 * b;
    } else {
        // signal filtering using identity a + b + c = 0. Assumes measurement error is normally distributed.
        float mid = (1.f/3) * (current.a + current.b + current.c);
        float a = current.a - mid;
        float b = current.b - mid;
        i_alpha = a;
        i_beta = _1_SQRT3 * a + _2_SQRT3 * b;
    }
```

This could be made a function getAlphaBeta or getClarkTransform that uses foc currents as inputs, and returns alpha and beta.
A function getParkTransform could use the electrical angle, alpha and beta and return DQcurrents.
getDCCurrent could use the electrical angle, alpha and beta and return DCcurrent.

Eventually, foc_current torque control type in loopfoc would do:
- getPhaseCurrents
- getClarkTransform
- getParkTransform
- PI control

And dc_current torque control type in loopfoc would do:
- getPhaseCurrents
- getClarkTransform
- getDCcurrent
- PI control

Candas1:
If this generates too much work (changing the examples, ....)
This alternative should have less impact:

getfoccurrents:
- getPhaseCurrents
- getClarkTransform
- getParkTransform

getdccurrents:
- getPhaseCurrents
- getClarkTransform
- calculate dc current

and no changes in loopfoc.

Candas1:
[...]

PR #340, opened 2023-11-19
Foc refactoring

Candas1:
This is the refactoring I mentioned about [here](https://github.com/simplefoc/Arduino-FOC/issues/311).
It prevents the code duplication in getDCcurrents and getfoccurrents, this will make future changes in the Clarke transform easier (e.g. using the best 2 phases)
getDQCurrents could be used when implementing FOC for stepper motors.

This shouldn't change how FOC works.

### DIFF
--- a/src/common/base_classes/CurrentSense.cpp
+++ b/src/common/base_classes/CurrentSense.cpp
@@ -39,12 +39,8 @@ DQCurrent_s CurrentSense::getFOCCurrents(float angle_el){
     ABCurrent_s ABcurrent = getABCurrents(current);
     
     // calculate park transform
-    float ct;
-    float st;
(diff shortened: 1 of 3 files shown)


runger1101001:
Hey, you didn't include speed comparison, but I assume you compared it? Is there a performance impact?

Looks great to me!

Candas1:
It is very similar:

Before:
RAM:   [=         ]   9.0% (used 4428 bytes from 49152 bytes)
Flash: [===       ]  26.8% (used 70280 bytes from 262144 bytes)
loopfoc=280us

After:
RAM:   [=         ]   9.0% (used 4428 bytes from 49152 bytes)
Flash: [===       ]  26.8% (used 70224 bytes from 262144 bytes)
loopfoc=280us

PR #341, opened 2023-11-19
Stm32 adc no interrupt

Candas1:
On stm32, when injected adc is used, the adc samples are saved in specific register.
The implementations for F1/F4/G4/L4 use an interrupt to save injected adc values in an array.
The way simplefoc works now, this is not required as those values are accessible anytime from the registers.

I introduced a SIMPLEFOC_STM32_ADC_INTERRUPT build flag.
By default, it will remove any code related to the adc interrupt.
When SIMPLEFOC_STM32_ADC_INTERRUPT is declared, code will work as before.
The reason why I kept this code as optional is that someone my want to use this interrupt for different reasons (e.g. running loopfoc, triggering an output for the scope).

I could test this on stm32f1 only, here are the results:
Without interrupt
RAM:   [=         ]   8.9% (used 4376 bytes from 49152 bytes)
Flash: [===       ]  26.7% (used 69916 bytes from 262144 bytes)
loopfoc = 270us

With interrupt
RAM:   [=         ]   9.0% (used 4428 bytes from 49152 bytes)
Flash: [===       ]  26.8% (used 70284 bytes from 262144 bytes)
loopfoc = 290us

It is slightly reducing the code size and reducing loopfoc duration as this interrupt was stopping the loopfoc execution several times unnecessarily. 
Because of the priority of this interrupt, it might also delay the hall sensor interrupt.

Additionally this PR has the adc calibration for G4/L4 ( this was already merged for F1 and G431-ESC1).

### DIFF
--- a/src/current_sense/hardware_specific/stm32/stm32f1/stm32f1_hal.cpp
+++ b/src/current_sense/hardware_specific/stm32/stm32f1/stm32f1_hal.cpp
@@ -124,9 +124,11 @@ int _adc_init(Stm32CurrentSenseParams* cs_params, const STM32DriverParams* drive
     HAL_ADCEx_InjectedConfigChannel(&hadc, &sConfigInjected);
   }
   
+  #ifdef SIMPLEFOC_STM32_ADC_INTERRUPT
   // enable interrupt
   HAL_NVIC_SetPriority(ADC1_2_IRQn, 0, 0);
   HAL_NVIC_EnableIRQ(ADC1_2_IRQn);
+  #endif
   
   cs_params->adc_handle = &hadc;
 
@@ -151,6 +153,7 @@ void _adc_gpio_init(Stm32CurrentSenseParams* cs_params, const int pinA, const in
 
 }
 
+#ifdef SIMPLEFOC_STM32_ADC_INTERRUPT
 extern "C" {
   void ADC1_2_IRQHandler(void)
   {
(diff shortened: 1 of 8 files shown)


runger1101001:
A 7% speed improvement as well!

I think in a few releases we can probably drop the interrupt code. I'm not aware that anyone is using it for anything...

Candas1:
If a chip can run loopfoc close to pwm frequency, it should have less impact.

Candas1:
Forgot that I am running the pwm at 'only' 16khz.
The higher the pwm frequency the more the adc interrupts would impact loopfoc

Candas1:
Loopfoc duration decreased from 360us to 240us on a STM32F1 if I compare the master branch and this PR 🚀​🚀​🚀​

runger1101001:
I think we'll make this part of the next release?

Candas1:
It's OK. This is a small performance improvement for a particular use case (stm32 - low side current sensing).
This release already has a lot of improvements.

Candas1:
I received a SimpleFOC so I hope I can do more tests now on F4 and L4.

I was thinking about this change again and I have a small doubt.
When not using the interrupt, it can be we read the data from the injected adc register while it's being written.
So phase A could be a new sample, and phase B could be from the previous sample.
But I think this can also happen with DMA without interrupt for the B431-ESC1 current sense driver.

I need to do more tests to check if it happens and what is the impact.

PR #343, opened 2023-11-25
fix some bugs in B-G431-ESC1 example

zhangyq-227:
Hi, I find some bugs in B-G432B-ESC1 example when I test.
1. In the loop function, we need a loopFOC() function to start the motor.
2. build_opt.h file need to add "lib_archive=false" to compile correctly in platformIO.
All changes are verified on my real hardware.
I‘m sorry, My English is not very good. I look forward to your reply. 
Wish you a happy life.

### DIFF
--- a/examples/hardware_specific_examples/B_G431B_ESC1/B_G431B_ESC1.ino
+++ b/examples/hardware_specific_examples/B_G431B_ESC1/B_G431B_ESC1.ino
@@ -5,13 +5,13 @@
 #include <SimpleFOC.h>
 
 // Motor instance
-BLDCMotor motor = BLDCMotor(11);
+BLDCMotor motor = BLDCMotor(12);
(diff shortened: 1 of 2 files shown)


runger1101001:
Thank you very much for reporting this! Your English is just fine, I understand you perfectly.

runger1101001:
I've changed this PR from main to dev branch, we only accept changes to the dev branch.

However, there are also some compilation problems with this PR - I think it doesn't like the lib_archive=false in the build_opts file. Normally this option should be put in your platformio.ini file...

Also your changes to the example file include changing the sensor type, using current sensing, etc... so I think we should restrict the changes just to the bug itself, if possible.

I think if it is ok with you and to save you further work I will manually introduce just the change with the missing loopFOC() in a separate commit.

runger1101001:
Done, in this PR: https://github.com/simplefoc/Arduino-FOC/pull/349

PR #346, opened 2023-11-26
Fix ESP32 Lowside Current Sense ADC2 init

mcells:
Fixes a bug, where lowside sensing using the ADC2 triggered the correct adc conversion, but then tried reading ADC1, resulting in a WDT error.


nmscode:
I have verified this fix on the ESP32 Pico Devkit M2 board that uses the ESP32-PICO-MINI-02.

runger1101001:
Thanks so much for finding and fixing this! ❤️

Issue #344, opened 2023-11-25
[BUG] Fix KV initialization


runger1101001:
like this
![image](https://github.com/simplefoc/Arduino-FOC/assets/7306964/229dc099-27cf-4ab5-b8b6-8ce3be814cf7)
can someone check my math make sure im not crazy lol

Change the sqrt(2) to sqrt(3) and it should be ok.
But maybe put the sqrt(3) somewhere else. It doesn't make sense to have a strange kv with the same name.

Elwin — Yesterday at 10:39 PM
Because the id iq vd and vq are like phase-zero quantities. The kv is in phase-phase.

Elwin — Yesterday at 10:42 PM
I think you should change it here
https://github.com/simplefoc/Arduino-FOC/blob/f9e9a2d29e912957bd75302d4f9a83a1cdff37a5/src/BLDCMotor.cpp#L399

Elwin — Yesterday at 10:47 PM
Is the code ment for motors with a trapezoidal back emf? Because if I recall correctly than it could be wrong what I just said. I only work with sinusoidal bemf motors.

runger1101001:
PR merged, so this will be closed with the 2.3.3 release

PR #347, opened 2023-11-27
Kv rating fix

nmscode:
fixes: #344 

Need to verify that this is the correct math/logic. It assumes sinusoidal bemf. Perhaps can add an option for more trapezoidal bemf motors.

### DIFF
--- a/src/StepperMotor.cpp
+++ b/src/StepperMotor.cpp
@@ -16,7 +16,7 @@ StepperMotor::StepperMotor(int pp, float _R, float _KV, float _inductance)
   phase_resistance = _R;
   // save back emf constant KV = 1/K_bemf
   // usually used rms
-  KV_rating = _KV*_SQRT2;
+  KV_rating = _KV;
   // save phase inductance
   phase_inductance = _inductance;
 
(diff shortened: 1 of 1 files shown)


runger1101001:
Thank you very much for contributing this!

I'm going to hold off merging it until after the next release, which I hope will be quite soon. That will give us some time to then test the effects on the dev branch.

Introducing this change will correct the calculations, but it will have the side effect of changing the behaviour of existing setups. Many people will have "tuned" their settings, picking a KV value for themselves which differs from the actual one but  works well in practice. People will have to change these values after we introduce the fix. So its something we need to do carefully.

runger1101001:
I have another question: why not pre-multiply the KV_rating value by the factor of _SQRT3/_RPM_TO_RADS in the init method, and then not have to do this multiplication/division each time through the loop?

nmscode:
That is also a possibility. I was not sure if the KV_rating variable might be used for something else and would be needed without the multiplier. However, I can make the change so that it is pre-multiplied if that is preferable.

Candas1:
I also have the feeling this will be confusing if KV_rating is used somewhere else.
And someone changing kv_rating after the constructor should know that as well.
It could be this is optimized by the compiler anyway.

runger1101001:
You're right - it would be confusing to have the parameter so named, but the value represent a different quantity.

So keep as is, but perhaps remove the unneeded brackets from the calculation - not sure if the compiler would respect the order implied by the parenthesis or optimize it away, but in this case we'd prefer the brackets around the two constants to make it clear to the compiler?

askuric:
This looks good to me. 
I've added sqrt(2) in order for the people to avoid setting too high KV values, to the KV values that people set relatively similar to the ones given by the manufacturers.
Honestly, I've found sqrt(2) to be empirically a good number. But I did not go any further with it.

This solution is much better both in terms of the code that is easy to understand(sqrt(3) instead sqrt(2)), and keeping the KV_rating the same as the one user has set will make it much easier to set for the user.

askuric (approved):

Candas1:
Hi Guys,

Is the statement in the docs  about increasing the kv by 10-20% still valid now ? 
https://docs.simplefoc.com/voltage_torque_mode#voltage-control-with-current-estimation-and-back-emf-compensation

image

PR #359, opened 2023-12-17
Prevent double constants

Candas1:
For constants as float instead of double.
Slightly reduces flash usage and loopfoc duration.

Before:
RAM:   [=         ]   9.0% (used 4400 bytes from 49152 bytes)
Flash: [===       ]  26.7% (used 69992 bytes from 262144 bytes)
loopfoc=260us

After:
RAM:   [=         ]   9.0% (used 4400 bytes from 49152 bytes)
Flash: [===       ]  26.6% (used 69856 bytes from 262144 bytes)
loopfoc=256us


runger1101001:
Merged it. Thank you!!

PR #365, opened 2024-01-07
Add currentsense.enable/disable

runger1101001:
Add a enable() and disable() function to currentsense, and reset the PIDs on a motor.enable().

See discussion here:
https://community.simplefoc.com/t/read-a-potentiometer-continuously-on-stm32-and-analogread/4214/90

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -111,6 +111,8 @@ void BLDCMotor::init() {
 // disable motor driver
 void BLDCMotor::disable()
 {
+  // disable the current sense
+  if(current_sense) current_sense->disable();
   // set zero to PWM
   driver->setPwm(0, 0, 0);
   // disable the driver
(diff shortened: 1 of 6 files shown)


Candas1:
Hi Richard,

I wanted to start implementing current sense enable/disable on STM32, but after I created a branch on my fork based on the project's dev, I end up with a **arduino-FOC** and a **Simple FOC** folder in the dependencies:
image

It looks like:
- https://github.com/simplefoc/Arduino-FOC.git#dev brings the Arduino-FOC folder
- https://github.com/simplefoc/Arduino-FOC-drivers.git#dev brings Simple FOC and SimpleFOCDrivers

Do you have the same issue ?
I updated platformio just before that, so it could be a bug there.

runger1101001:
Is maybe one of them the original project dependency?

SimpleFOC is the project and library name, but Arduino-FOC is the actual folder/repository name. At a guess I'd say one of them comes from a git pull, and the other via library dependency or symlink?

I can't say I have the same issue, and when I check the dev branch online I can't see any trace of SimpleFOC within the drivers?

Candas1:
ok I won't waste too much time for now.
the arduino-foc folder has your last changes and points to my fork, so it should be good enough.

So just to understand, how should I declare the enable and disable function ?
Should I override it in lowsidecurrentsense class, and call new _enable() and _disable() functions in the current sense driver ?

PR #367, opened 2024-01-14
LinearHall improvements and move to main repository

dekutree64:
Changes compared to the original pull request in the drivers repository https://github.com/simplefoc/Arduino-FOC-drivers/pull/12
1. Added a version of init which turns the motor one revolution to find the center values of the sensors.
2. Moved the calls to analogRead into a weakly bound function ReadLinearHalls so it can be overridden with custom ADC code on platforms with poor analogRead performance.
3. Commented out the pinMode calls in init, which makes it possible to pass in ADC channel numbers for custom ReadLinearHalls to use without having to remap them every update.
4. Changed to use the much faster _atan2 function that was added to foc_utils recently.
5. Added examples.

### DIFF
--- a/examples/motion_control/position_motion_control/linear_hall/angle_control/angle_control.ino
+++ b/examples/motion_control/position_motion_control/linear_hall/angle_control/angle_control.ino
@@ -1,0 +1,117 @@
+/**
+ *
+ * Position/angle motion control example
+ * Steps:
+ * 1) Configure the motor and hall sensor
+ * 2) Run the code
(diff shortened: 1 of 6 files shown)


runger1101001:
Hey, from my point of view I think we should add the LinearHall code in the drivers repository, and not the main repository...
While it's very cool, at the moment it's something that only very few of our users will be able to use.

dekutree64:
Closing this and opening a new PR in drivers repository

PR #368, opened 2024-01-20
expose result of PP check

runger1101001:
Expose result of PP check to enable firmware to check for this.


runger1101001:
Tested on RP2040

PR #369, opened 2024-01-21
RP2040: respect pin nrs on boards which remap them

runger1101001:
Some boards like to remap the pin numbers so they match the Arduino convention rather than following the RP2040 GPIO convention.
This patch restores compatibility with those boards.

In particular, this makes SimpleFOC work with the Arduino Nano RP2040 Connect.


runger1101001:
Tested on Arduino Nano RP2040 Connect

PR #371, opened 2024-01-26
[BUGFIX] Injected channel

askuric:
There seems to be a bug in the code  ( one pretty hard to find, believe me :D )

Basically when in the for loop the injected channels are read the current code  does a completely reasonable assumption that
`ADC_INJECTED_RANK_1=1`, `ADC_INJECTED_RANK_2=2`, and `ADC_INJECTED_RANK_3=3`, 
However, at least in case of stm32g474 this is not the case :D 
So here is the fix to avoid this assumption.

### DIFF
--- a/src/current_sense/hardware_specific/stm32/stm32f1/stm32f1_mcu.cpp
+++ b/src/current_sense/hardware_specific/stm32/stm32f1/stm32f1_mcu.cpp
@@ -88,8 +88,10 @@ float _readADCVoltageLowSide(const int pin, const void* cs_params){
     if( pin == ((Stm32CurrentSenseParams*)cs_params)->pins[i]){ // found in the buffer
       #ifdef SIMPLEFOC_STM32_ADC_INTERRUPT
         return adc_val[_adcToIndex(((Stm32CurrentSenseParams*)cs_params)->adc_handle)][i] * ((Stm32CurrentSenseParams*)cs_params)->adc_voltage_conv;
-      #else
-        return HAL_ADCEx_InjectedGetValue(((Stm32CurrentSenseParams*)cs_params)->adc_handle,i+1) * ((Stm32CurrentSenseParams*)cs_params)->adc_voltage_conv;
+      #else 
+        // an optimized way to go from i to the channel i=0 -> channel 1, i=1 -> channel 2, i=2 -> channel 3
+        uint32_t channel = (i == 0) ? ADC_INJECTED_RANK_1 : (i == 1) ? ADC_INJECTED_RANK_2 : ADC_INJECTED_RANK_3;;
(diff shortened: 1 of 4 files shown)


Candas1:
Hi,

I added this code for using injected adc registers and not using the interrupt anymore.
For the channel I use i+1, it wasn't OK ?

askuric:
Your addition works great it's just that for some architectures the `i+1` did not correspond to the `ADC_INJECTED_RANK_x`. 
In particular my nucleo stm32f474re, in the case fo G4 architecture it is defined using LL flags 
```cpp
#define ADC_INJECTED_RANK_1                (LL_ADC_INJ_RANK_1) /*!< ADC group injected sequencer rank 1 */
```
here is the [link](
https://github.com/stm32duino/Arduino_Core_STM32/blob/50629bd31f2edc873363f0be13ce012c6974d9ff/system/Drivers/STM32MP1xx_HAL_Driver/Inc/stm32mp1xx_hal_adc_ex.h#L281)

For some architectures it is just from 1 to 3 but for others its done using these LL flags. [stm32duino repo search](https://github.com/search?q=define+ADC_INJECTED_RANK_1&type=code). I am not even sure that these LL flags are consecutive.  So it's safer to enforce using the  `ADC_INJECTED_RANK_x`.

Candas1:
My bad, first bug.

Candas1:
What driver are you using with your nucleo board that has 3 shunts ?
I need a proper 3 shunt setup for my tests.

runger1101001:
Here's a higher power one with 3 shunt sensing: https://www.aliexpress.com/item/4000126430773.html
It also has 2 shunt sensing for some reason (as someone recently pointed out on the forum) so this is even a board where we could run 2x current sensing at the same time :-D

This one: https://www.st.com/resource/en/data_brief/x-nucleo-ihm07m1.pdf
Will plug directly into a Nucleo64, and has 3-shunt sensing. Not so high power, but might be convenient...

I also have a DRV8316 based board which has current sensing (3 shunt low side), I could send you one if you want :-) but it is in the shape of a Arduino MKR Shield so you'd need jumper wires to connect it to STM32s.

Candas1:
Actually I already have:
[X-NUCLEO-IHM07M1](https://www.st.com/en/ecosystems/x-nucleo-ihm07m1.html)
[X-NUCLEO-IHM16M1](https://www.st.com/en/ecosystems/x-nucleo-ihm16m1.html)
An even managed to order this one when it was finally in stock:
[X-NUCLEO-IHM16M1](https://www.st.com/en/ecosystems/x-nucleo-ihm08m1.html)

I need to go through the docs and see if I can use all shunts on same ADC with the nucleo boards I have.

askuric:
Merging into the dev branch.

PR #373, opened 2024-01-26
Fix bug in ATMega sine computation

runger1101001:
fix floating point constant incorrectly converted to int


runger1101001:
Tested on Arduino Nano (original)

PR #384, opened 2024-02-18
Add a MotionControlMode::angle_nocascade which uses a single PID for angle control

runger1101001:
As discussed here: https://community.simplefoc.com/t/motor-oscillates-when-in-cascaded-position-velocity-control-loop/4194

I've tested this and it works quite nicely. I imagine some people might find it useful.


Copper280z:
I think this would be a useful change.

Issue #166, opened 2022-04-03
STM32 PWM generation on TIM14/TIM15/TIM16 has double frequency


runger1101001:
TIM14/TIM15/TIM16 seem to be a special case, maybe some of the other high-number timers too. They wind up with double the frequency - so the PWM is correct, but with double frequency to the other timers.

This bug exists on both master and dev branches (i.e. in both original and refactored STM32 drivers)

Here the result of:
`BLDCDriver3PWM driver = BLDCDriver3PWM(PB8, PA10, PB3);`
on a STM32L433: 
![image](https://user-images.githubusercontent.com/7306964/161452098-ae8c240f-432c-4322-9b35-41239b8a1754.png)
Note that phase A is the top channel, with 20µs period, which is on TIM16. The other two phases are on TIM1 and TIM2.

I had similar results with TIM14 on STM32G031 and TIM15 on STM32G474, so I think it has to do with these timers.

runger1101001:
TBH I would have to retest. Some STM32s have different clock domains for different timers, something I did not appreciate when I created this issue. So I think these high number timers were fed from a different clock path, which presumably had a different divider set in the L433 board I was using.

Now we have a stm32_getTimerClockFreq(TIM_HandleTypeDef *handle) which takes this into account, and is used to set the timer frequency.

So I assume that it is solved, but I have not tested it.

PR #388, opened 2024-03-11
Set STM32 timer prescaler based on timer clock frequency

Copper280z:
This commit adds a function which calculates the correct timer prescaler value based on the actual timer clock frequency and sets it. This is necessary because some stm32duino clock configs, specifically the nucleo-f746zg, set things up such that the timers have an input frequency that differs by 2x. Families other than F7 also allow for this, but don't seem to have the same problem with the default configs.
image

HardwareTimer->setOverflow(num, HERTZ_FORMAT) attempts to deal with this, but it does so by setting TIMx->ARR to a value that differs by 2x, instead of setting the prescaler. For some PWM frequencies this leads to an off-by-one error in ARR between the different timers(with the 2x factor taken into account). The values I observed with the aforementioned f746 were 

```
TIM1->ARR = 1349 (2698)
TIM4->ARR = 2699
```

This new function finds the lowest timer frequency for all timers, then calculates the prescaler value for each timer that's required to make the counter frequency match. It also calculates the correct overflow value for this frequency and sets it with HT->setOverflow(num, TICK_FORMAT), which does not reset the prescaler value. This method does require that no further calls to setOverflow are made with HERTZ_FORMAT, otherwise the prescaler value will be overwritten.

Additionally this change also adds the _getPwmState function, in anticipation of merging HFI support, as well as adding the STM32F7 family to the list of defines in _getInternalSourceTrigger. The F7 family timer interconnections appear to mirror the F4 family.
pg713 of RM0385 rev8 (STM32F74xx and F75xx reference manual)
image
pg794
image

So far I've tested this on the F746 nucleo + SFOC shield and it works as expected. I haven't tested further on other targets.

### DIFF
--- a/src/drivers/hardware_specific/stm32/stm32_mcu.cpp
+++ b/src/drivers/hardware_specific/stm32/stm32_mcu.cpp
@@ -26,7 +26,15 @@ PinMap* timerPinsUsed[SIMPLEFOC_STM32_MAX_PINTIMERSUSED];
 
 
 
+bool _getPwmState(void* params) {
+  // assume timers are synchronized and that there's at least one timer
+  HardwareTimer* pHT = ((STM32DriverParams*)params)->timers[0];
+  TIM_HandleTypeDef* htim = pHT->getHandle();
+  
+  bool dir = __HAL_TIM_IS_TIM_COUNTING_DOWN(htim);
 
+  return dir;
+}
 
 
 // setting pwm to hardware pin - instead analogWrite()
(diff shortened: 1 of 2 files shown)


runger1101001:
Hey, this may also explain this issue? https://github.com/simplefoc/Arduino-FOC/issues/166

May I ask, is this tested in combination with Antun's new timer-synchronisation based on triggers?

Copper280z:
I think it might explain #166, at least based on a quick read. I'd need to check the various docs to confirm though. I'd expect this to resolve it, in any case, so long as HardwareTimer handles those timers correctly.

Yes, this is tested with Antun's timer sync code, which, to be clear, only syncs the start of the timers. It does not do any synchronization once they're running, that requires reset mode along with an update trigger, instead of an enable trigger. Using update/reset mode breaks current sensing in a difficult to fix way, at least without this change. I didn't test it with this change.

I ran into this whole trying to port current sensing and HFI to the F7, which strongly depend on having sync'd timers. I found that the effected timer was not sync'd even though it had nominally the same frequency on the scope.

askuric:
Thanks for this. LGTM.

askuric:
Hi @Copper280z,
Did you maybe use f7 board with low-side current sensing or was it always inline?
I was wondering if you know is some of the architectures f4/l4/g4 has the same low-side ADC configuration as f7.

Copper280z:
I used it with lowside current sensing. 

https://github.com/simplefoc/Arduino-FOC/pull/394#issue-2254582517

PR #389, opened 2024-03-19
Add missing implementation for MagneticSensorI2C::AS5600

ystradmann:
The current declaration of `MagneticSensorI2C` already has a public static member `AS5600()`, which is however lacking an implementation. This change adds the (trivial) implementation.


askuric:
Thanks, this discrepancy is in the code for a loong time :D
LGTM

PR #394, opened 2024-04-20
Add lowside current sensing for STM32F7 microcontrollers

Copper280z:
This PR adds lowside current sensing support for the F7 family of chips. The F7 and F4 are very, very similar and could likely be merged to a single set of files with only a handful of inline #ifdefs. 

I think the largest difference is that the F7 supports more timer TRGO lines than the F4 and that the F7 has a silicon bug that sometimes requires enabling other clock sources besides the ADC, or setting the trigger to use both edges instead of just rising.
```
  sConfigInjected.ExternalTrigInjecConvEdge = ADC_EXTERNALTRIGINJECCONVEDGE_RISINGFALLING;  
```

```    
if (!IS_TIM_REPETITION_COUNTER_INSTANCE(instance_to_check)) {
      // workaround for errata 2.2.1 in ES0290 Rev 7
      // https://www.st.com/resource/en/errata_sheet/es0290-stm32f74xxx-and-stm32f75xxx-device-limitations-stmicroelectronics.pdf
      __HAL_RCC_DAC_CLK_ENABLE();
    } 
```

### DIFF
--- a/src/current_sense/hardware_specific/stm32/stm32f7/stm32f7_hal.cpp
+++ b/src/current_sense/hardware_specific/stm32/stm32f7/stm32f7_hal.cpp
@@ -1,0 +1,185 @@
+#include "stm32f7_hal.h"
+
+#if defined(STM32F7xx)
+
+//#define SIMPLEFOC_STM32_DEBUG
+
+#include "../../../../communication/SimpleFOCDebug.h"
+#define _TRGO_NOT_AVAILABLE 12345
+
+ADC_HandleTypeDef hadc;
+
+int _adc_init(Stm32CurrentSenseParams* cs_params, const STM32DriverParams* driver_params)
+{
+  ADC_InjectionConfTypeDef sConfigInjected;
+
+  // check if all pins belong to the same ADC
+  ADC_TypeDef* adc_pin1 = (ADC_TypeDef*)pinmap_peripheral(analogInputToPinName(cs_params->pins[0]), PinMap_ADC);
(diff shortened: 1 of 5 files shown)


askuric:
Oh Wow!
I've just thrown the comment in the other PR, just in case you had some ideas, and you've come with the complete solution :D
Awesome,  thanks a lot for this. I'll just add a compiler check in the github CI for f7 architecture and I'm gonna merge it to the upcoming release. 
Even though it is not that well tested from our part, I think its a good way to get it to the final users and get some feedback. We will continue testing it in the dev branch.

Copper280z:
I actually wrote this some time ago, but merging it wouldn't make sense until the timer issue was resolved, because the easiest test case of a nucleo+sfoc shield wouldn't work without the prescaler fix. I have several more nucleos that I'll slowly work on adding current sense support for as well. H7 will probably be the next as it's well supported by the tools.

With the number of families of chips, it might make sense to consider reworking the stm32 current sensing backend in a way similar to how @Candas1 has been working on. This would reduce the amount of almost exactly repeated code.

For this pr, since I was testing on an sfoc shield with inline sensors, the biggest risk to mitigate would be that I mixed up what pwm phase, v0 or v7, to sample the current on. If anyone has the hardware to conclusively test this with lowside sensing, that would probably be a good thing. The 2 situations to test would be where the timer that's used for the adc triggering has a repetition counter and where it doesn't, since the repetition counter is used instead of an ISR if available.

PR #407, opened 2024-06-07
Added gain documentation to B_G431B_ESC1.ino

rob-deutsch:
I spent quite a while trying to find how this gain was calculated, and even asked on Discord. Hopefuly this documentation addition will save the next person some time.


runger1101001:
Thanks for this!

Issue #402, opened 2024-05-21
MagneticSensorI2C uses incorrect bit mask calculation


runger1101001:
https://github.com/simplefoc/Arduino-FOC/blame/6fafa35b73e9c5ba6f5d9259640a5ed328dc2750/src/sensors/MagneticSensorI2C.cpp#L40

```
  lsb_mask = (uint8_t)( (2 << lsb_used) - 1 );
  msb_mask = (uint8_t)( (2 << bits_used_msb) - 1 );
```

Why are we shifting 2 here? It does not seem correct.

Some examples: 2 is 00000010b in binary

lets say lsb_used is 6, i.e. we would want bits 7..2 of the LSB, with bits 0 and 1 not used. But this calculation would give us:
00000010b << 6 = 10000000b
10000000b -1 = 01111111b and we get bits 6..0 instead

lets say msb_used is 4, i.e. we would want bits 3..0 of the MSB, but the calculation gives us:
00000010b << 4 = 00100000b
00100000b -1 = 00011111b and we get bits 4..0 instead

This is just wrong. It was probably never noticed for two reasons. Firstly, the present calculation, though incorrect, still gives the correct mask for the LSB if all 8 bits are used. Secondly, although the MSB mask is wrong, it appears the AS5600 actually always returns 0 for the incorrectly included bits, and so no one has ever really noticed this...

The correct calculation should be:
```
lsb_mask = (uint8_t)( ( (0x1 << lsb_used) - 1 )  << (8-lsb_used) );
msb_mask = (uint8_t)( (0x1 << bits_used_msb) - 1 );
```

askuric:
Oh wow!
This is a big one. 
I don't remember what was my thinking any more. :)
This was written at the very early stages of simplefoc :D
The interesting part is that it has never been noticed though!

askuric:
Hey Richard, 

I've looked into it a bit and here is the structure of AS5600 :
![image](https://github.com/simplefoc/Arduino-FOC/assets/36178713/38b80a44-6756-4fad-bb67-d680e4a19330)
[datasheet page 19](https://look.ams-osram.com/m/7059eac7531a86fd/original/AS5600-DS000365.pdf)

And here is the AS5048B
![image](https://github.com/simplefoc/Arduino-FOC/assets/36178713/7af5724e-48db-4ce9-9505-be24eeae1ec2)
[...]

PR #412, opened 2024-06-14
Fix for MagneticSensorI2C uses incorrect bit mask calculation #402

runger1101001:
Update to MagneticSensor to handle bit mask correctly. This implementation should work with AS5600, AS5048B and MT6701 sensors.


runger1101001:
Still testing...

PR #414, opened 2024-06-22
408 feature support for arduino esp32 v301 current sensing

askuric:
Ok so this is the final pull request that unifies all the changes necessary for the API transition from v2.x to v3.x arduino-esp32. 

Main changes:
- MCWPM driver rewritten completely
   - made more flexible, no more slots
- Low-side current sensing rewritten completely for the new API
   - new timer synchronization code
   - made more flexible, no more hard-coded funcitons and variables 
   - could support more than 2 motors (but not suggested - adc read time too long)
- ADC fast reading rewritten 
   - simplified both for esp32 and esp32s variants
   - Now used by default both for mcwpm and ledc
- LEDC driver rewritten for the new API
   - Extended and made more flexible 
   - now center aligned
   - now supports 6pwm
- More elaborate error reporting and debugging scheme, easier to know what went wrong

Other new features:
- I've also mad the function `_driverSyncLowSide` return `-1` if it fails, so there are a few slight changes other mcu types, not just esp32
- I've made the functions `_configureLowSide` and `_readADCVolatgeLowSide`  return `-1`  (and display a debugging error) by default instead of calling their inline equivalents. This is a safety measure to avoid using suing low-side current sensing if it's not available.
   - this required a slight change in the BG431 driver

The code was tested on esp32 and esp32s3 variants, so it is relatively mature. 
It needs much more testing, but I think that it should be fine for the dev branch.


askuric:
We're going for the merge! 🎉  :D

Issue #415, opened 2024-06-29
[BUG] sin/cos integer overflow on 16-bit CPUs


dekutree64:
Thanks to ChrisMicro on the forum for finding this bug in the fast sin/cos added last year https://github.com/simplefoc/Arduino-FOC/pull/285

Forum thread https://community.simplefoc.com/t/simplefoc-sin-calculation/5131

There are two potential solutions:
Solution 1 is to change the lookup table to signed 16-bit and fudge the last entry to 32767 so it doesn't overflow, and add some casts to do part of the interpolation calculation in 32-bit:
return (1.0f/32768.0f) * (t1 + (int)((((int32_t)t2 - t1) * frac) >> 8));

Solution 2 is to change t1 and t2 to int32_t. This will be a few cycles slower on 16-bit platforms, but retain the behavior of returning exactly 1.0 at pi/2 and -1.0 at 3pi/2.

I would be fine with either. I think 1 was my original intent, but after all the iterations testing the speed and accuracy of the calculations, I forgot to make the final edits.

EDIT: No measured speed difference when testing, so I went with solution 2. Pull request #416

ChrisMicro:
The maximum value of an int16 is 32767. For this reason I would expect for a int16 sin calculation the maximum value in the table should be this one. To achieve the highest accuracy the whole table should be recalculated for this value.

dekutree64:
The table is unsigned, so it can go up to 65535. I went with 32768 so the quadrant points would return exactly 1.0 and -1.0, but thinking about it now, the return has a float multiplier so it could use (1.0f/65535.0f) and return the correct amplitude anyway.

This is the table generation code: for(int i = 0; i <= 64; i++) fprintf(file, "%i,", (int)(sin((i * M_PI/2) / 64) * 32768 + 0.5));

And here is the result with the constant changed to 65535:
0,1608,3216,4821,6424,8022,9616,11204,12785,14359,15924,17479,19024,20557,22078,23586,25079,26557,28020,29465,30893,32302,33692,35061,36409,37736,39039,40319,41575,42806,44011,45189,46340,47464,48558,49624,50659,51664,52638,53580,54490,55367,56211,57021,57797,58537,59243,59913,60546,61144,61704,62227,62713,63161,63571,63943,64276,64570,64826,65042,65219,65357,65456,65515,65535

[...]

PR #416, opened 2024-06-29
Fix for #415 sin/cos integer overflow on 16-bit CPUs

dekutree64:
Terrible bug in the fast sin/cos added last year, resulting in rough sine waves and a spike from +1 to -1 at pi/2


kalvdans (approved):

runger1101001:
Merging this before the next release.

Issue #418, opened 2024-07-06
[FEATURE] Hall sensor without interrupt


Candas1:
I wrote [this](https://github.com/Candas1/Arduino-FOC/commit/d7d44887737f6dd949c2fc06bb7a153d9b026b36) code some time ago but I haven't published it because I wasn't sure it was useful enough. I am posting it here to get your opinion, specially you @dekutree64  as you did a lot of work on this sensor.

What it does is:
- if no interrupts are enabled for hall sensors, it will just check the hall states directly when sensor is updated.
- some small code optimizations

The user can choose if he wants to run it with or without interrupts.

Benefits:
- if user forgets to enable interrupts, the motor will still run (we could issue a warning in debug mode)
- if there are not enough interrupts available (arduino?), the motor can still run
- With interrupts, the faster the motor spins the more interrupts are triggered, slowing down the loop duration/frequency. Without interrupts, the code runs at constant duration/frequency
- PIDs/filters might not be impacted by hall interrupts running in the middle
- less problems when there are noise issues ? (to be checked)

Downside:
- the slower the MCU is and the faster the motor spins, the less accurate the position will be

dekutree64:
Looks good to me.

I don't think you can drive faster than one update per sector anyway, since the lagging stator field will start decelerating the rotor instead of accelerating it.

You have to miss at least 2 pulses before the position is lost, so it would require some fast backdriving before it's a problem. The main place I can see that happening is rolling down a hill on an electric skateboard, and that doesn't need to keep track of absolute position.

The main thing you'll lose is accurate timestamps for velocity calculation. May need higher lowpass Tf to average it out, resulting in slower velocity PID response time. Probably not enough to be a problem, though.

Candas1:
Thanks a lot for your very valid comments.
I am getting 320 kv motors with hall sensors soon, I can test this at high speeds with slow and fast MCUs and compare the velocity with and without interrupts.

This is more like a fallback mode but I want to make sure it works.

dekutree64:
This is awesome. I can use 41F halls now, which don't need any filtering capacitors. They generate excessive interrupts even if I do add capacitors, but work effortlessly with this new code.

[...]

PR #424, opened 2024-07-17
Hall sensor without interrupt

Candas1:
This code will allow running a hall sensor without interrupt if the interrupts if enableinterrupts() is not used.
It will poll the hall sensor states during the sensor.update instead.
There are also small optimizations of the hall sensor updatestate code.

More details in [this](https://github.com/simplefoc/Arduino-FOC/issues/418) issue.

### DIFF
--- a/src/sensors/HallSensor.cpp
+++ b/src/sensors/HallSensor.cpp
@@ -42,22 +42,21 @@ void HallSensor::handleC() {
  * Updates the state and sector following an interrupt
  */
 void HallSensor::updateState() {
-  long new_pulse_timestamp = _micros();
-
   int8_t new_hall_state = C_active + (B_active << 1) + (A_active << 2);
 
   // glitch avoidance #1 - sometimes we get an interrupt but pins haven't changed
-  if (new_hall_state == hall_state) {
-    return;
-  }
+  if (new_hall_state == hall_state) return;
+
+  long new_pulse_timestamp = _micros();
   hall_state = new_hall_state;
 
   int8_t new_electric_sector = ELECTRIC_SECTORS[hall_state];
-  if (new_electric_sector - electric_sector > 3) {
+  int8_t electric_sector_dif = new_electric_sector - electric_sector;
+  if (electric_sector_dif > 3) {
     //underflow
     direction = Direction::CCW;
     electric_rotations += direction;
-  } else if (new_electric_sector - electric_sector < (-3)) {
+  } else if (electric_sector_dif < (-3)) {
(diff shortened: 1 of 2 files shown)


askuric:
I am in for merging this. for the v2.3.4! 
I'll integrate it to the docs as well.

askuric:
Hey @Candas1,

Would you have time to add an exmaple code in the library examples folder of how to use this new `HallSensor`. :D
I think the ideal would be to add it in the `examples/utils/sensor_test` folder:
https://github.com/simplefoc/Arduino-FOC/tree/master/examples/utils/sensor_test

Candas1:
Hi  @askuric,

It will just work by commenting [enableinterrupt()](https://github.com/simplefoc/Arduino-FOC/blob/463eb31f354dd80c29f0ae4f16d0c1733f208a0a/examples/utils/sensor_test/hall_sensors/hall_sensor_example/hall_sensor_example.ino#L33).

So it might not require an additional example, but I can add a comment there. Most important might be to explain in the docs when to use it.

Looking at the folder you shared, I have one doubt. Is [this](https://github.com/simplefoc/Arduino-FOC/blob/master/examples/utils/sensor_test/hall_sensors/hall_sensor_software_interrupts_example/hall_sensor_software_interrupts_example.ino) still required ?
- it works only on arduino's
- it requires an additional library
- as it runs each listener one by one, it can potentially run updateState() several times and with partial information if more than one hall state changed
- my code needs sensor.update() to run often enough, pcimanager seems to rely on [global interrupts](https://github.com/prampec/arduino-pcimanager/blob/32052dedd25cdbfe29069d6c504f9cf10197f3d2/src/PciManager.cpp#L154)

askuric:
Ok cool!
I'd still add an example to make sure that peole don't get ocnfused, but I can do it quickly as its so simple, don't worry about it. 

Yeah, I'd still keep the software interrupts for now, we can remove them in the future releases. 

I'm updating the docs at the moment. Would you agree with this statement?

**RULE OF THUMB: Hardware/Software or No interrupts?**
- Start with the no interrupts and see if the performance is good enough for your application.
- If the performance is not good enough and if you have enough hardware interrupt pins, try using hardware interrupts.
- Otherwise, if you are using Arduino boards, try using software interrupts. (worst performance solution to be expected)

Candas1:
Yes that's ok.

We will get more feedback in the future as users experiment, so we can add more details later if needed. (fast vs slow hardware, fast vs slow motor, accuracy for velocity control, ....)

PR #426, opened 2024-07-21
Current PI Autotuning

Candas1:
Shared by @mcells from his fork https://github.com/mcells/Arduino-FOC/blob/d40c7d7fddff4c3e58b26ebd2e2cc846fddfc09e/src/HFIBLDCMotor.cpp#L71 

### if phase resistance and phase inductance are known:
Will initialize Kp and Ki for q and d current PI controllers automatically based on phase resistance, phase inductance and current bandwidth (default 300).
Needs to be adapted if Ld and Lq are available in the future.
### if not known:
Same SimpleFOC default values as before are used.
### users that have tuned their PIs
They will overwrite Kp and Ki from their code.

What if users have set phase resistance and phase inductance but were using default PI parameters ? Is their an impact ?

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -82,7 +82,16 @@ int BLDCMotor::init() {
     // current control loop controls voltage
     PID_current_q.limit = voltage_limit;
     PID_current_d.limit = voltage_limit;
+
+    if (_isset(phase_resistance) && _isset(phase_inductance)){
+      PID_current_d.P = phase_inductance * DEF_CURR_BANDWIDTH * _2PI;
+      PID_current_d.I = PID_current_d.P * phase_resistance / phase_inductance;
+      
+      PID_current_q.P = phase_inductance * DEF_CURR_BANDWIDTH * _2PI;
+      PID_current_q.I = PID_current_q.P * phase_resistance /phase_inductance;
+    }
   }
+  
   if(_isset(phase_resistance) || torque_controller != TorqueControlType::voltage){
     // velocity control loop controls current
     PID_velocity.limit = current_limit;
--- a/src/StepperMotor.cpp
+++ b/src/StepperMotor.cpp
@@ -48,6 +48,21 @@ int StepperMotor::init() {
   if(voltage_sensor_align > voltage_limit) voltage_sensor_align = voltage_limit;
 
   // update the controller limits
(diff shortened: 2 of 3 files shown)


mcells:
Great!
One thing though, that part of the code is not by me, all credit goes to @Copper280z. I just had my fork on hand quicker when sharing this earlier🙈
Thanks for the mention regardless!

Copper280z:
To be clear, this method is really more like pole placement, rather than autotuning. 

I would recommend also setting the filter cutoff based on the controller bandwidth, otherwise the controller will have a higher bandwidth than the filter likely leading to some integral windup or something else undesirable happening. The default filter for faster chips has a cutoff frequency of ~30hz. Something like 1.5-5x the controller bandwidth (so 450-1500hz) is probably a good place to start. In the hfi fork we've been using a fixed ~2khz lowpass for both controllers.

askuric:
Oohh, very very very nice!
Do you guys have some refereces/papers about this?

Candas1:
Some time ago I shared this:
https://community.simplefoc.com/t/measuring-the-resistance-of-the-motor/3647/11?u=candas1

askuric:
Nice, I'll look into it.

Igitigit2:
Is it intended that q and d PI gains are the same, or would it be better to use inductance_q and inductance_d instead of the phase inductance (if you have them...)?

Copper280z:
If the inductance for each axis is available then using it would be more accurate, but using the average is ok if that info isn't an option.

uLipe:
> Oohh, very very very nice!
> Do you guys have some refereces/papers about this?

I just went to the code, the tunning procedure is just the result of the compensator for the 1st order model formed by the motor current loop, when you model the motor assuming the non-linearities are not a problem you will end up into the inner loop transfer function which has just one pole to compensate for tracking error and response time, with motor inductance and resistance being the parameters that determines the what is the response time (for the P component) and how much of the tracking error you would like to compensate (for the I term).

There is a ancient topic in the C2000 from TI where they cover a lot of practical details behind the values:
https://e2e.ti.com/support/microcontrollers/c2000-microcontrollers-group/c2000/f/c2000-microcontrollers-forum/1060236/tms320f28035-calculating-theoretical-values-for-current-loop-pi-controller-for-motor-control

PR #435, opened 2024-10-03
Add RP2350 support

runger1101001:
Since our driver is based on the Pico SDK functions, and these were not changed for the RP2350, and the timer functionality is more or less the same, it looks like there is not much to do to support the Pico 2.

It can be tested in platformio using Earle Phil hower's core and Max Gerhard's PlatformIO support, with a configuration for platformio.ini like:

```
[env:pico]
platform = https://github.com/maxgerhardt/platform-raspberrypi.git
board = rpipico2
framework = arduino
board_build.core = earlephilhower

lib_archive  = false
```


runger1101001:
Tested using RPi Pico 2 and Earle Phil Hower core.

PR #439, opened 2024-10-14
Silabs MGM240S (Gecko) support

runger1101001:
PWM driver support for SiLabs Gecko type MCUs as used in the Arduino Nano Matter

Currently supports:
- 3 PWM (tested)
- 1 PWM, 2 PWM, 4 PWM (implemented but not tested)

Not yet implemented:
- 6 PWM

6 PWM support will be added, this MCU can support hardware 6 PWM with DTI for any pin combination :-)

In theory, 5 motors can be supported, in practice the MCU won't have enough pins or CPU muscle for so many, but maybe 2 could run ok.

In theory, current sensing should be possible, also low side sensing, based on DMA support and the peripheral reflex system (inter-peripheral events). I'll leave that as an exercise for another time :-)

Performance tests in closed loop mode to follow.

### DIFF
--- a/src/drivers/hardware_specific/silabs/silabs_mcu.cpp
+++ b/src/drivers/hardware_specific/silabs/silabs_mcu.cpp
@@ -1,0 +1,143 @@
+
+#include "./silabs_mcu.h"
+
+
+#if defined(ARDUINO_ARCH_SILABS)
+
+#include "em_device.h"
+#include "em_chip.h"
+#include "em_cmu.h"
+#include "em_emu.h"
+#include "em_gpio.h"
+#include "em_timer.h"
+
+#pragma message("")
+#pragma message("SimpleFOC: compiling for SiliconLabs")
+#pragma message("")
+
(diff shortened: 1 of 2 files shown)


runger1101001:
6-PWM support added:

image

Up to 5 motors with 6-PWM would be theoretically supported by the timer hardware. in practice the modules won't have enough pins or CPU power for it.

If using just one or two motors, then *any* pins can be used for hardware 6-PWM, which is nice and makes this MCU easy to use :-)

It also means I won't bother implementing software 6-PWM.

Dead-Time insertion is possible up to 64 timer counts on rising and falling side - assuming a prescaler of 0 on a 72MHz timer clock, and PWM frequency of 25kHz, then 2% dead time work out to 31 timer counts, and 400ns of dead time on each side. You could go as high as 4% (= 800ns) at 25kHz PWM.

Other notes:

- setting polarity active-high or active-low is implemented, also separately for 6-PWM. TODO test it
- phase state setting is not implemented in 6-PWM mode TODO implement and test it
- TODO add a test case to GitHub actions for Silabs
- TODO add documentation in docs.simplefoc.com

Issue #440, opened 2024-11-05
[BUG] MCPWM on ESP32-C6 with two BLDC motors


kondor1622:
I am trying to run two BLDC gimbal motors on ESP32-C6. 
When using only one of the motors everything works fine. 

The problem is with MCPWM setup when using second motor - error log:

```
============ Before Setup End ============
[   517][V][esp32-hal-uart.c:408] uartBegin(): UART0 baud(115200) Mode(800001c) rxPin(17) txPin(16)
[   526][V][esp32-hal-uart.c:497] uartBegin(): UART0 not installed. Starting installation
[   534][V][esp32-hal-uart.c:560] uartBegin(): UART0 initialization done.
[   541][V][esp32-hal-periman.c:235] perimanSetBusDeinit(): Deinit function for type GPIO (1) successfully set to 0x42004b6c
[   552][V][esp32-hal-periman.c:160] perimanSetPinBus(): Pin 7 successfully set to type GPIO (1) with bus 0x8
[   562][V][esp32-hal-periman.c:235] perimanSetBusDeinit(): Deinit function for type GPIO (1) successfully set to 0x42004b6c
[   573][V][esp32-hal-periman.c:160] perimanSetPinBus(): Pin 0 successfully set to type GPIO (1) with bus 0x1
[   582][V][esp32-hal-periman.c:235] perimanSetBusDeinit(): Deinit function for type GPIO (1) successfully set to 0x42004b6c
[   593][V][esp32-hal-periman.c:160] perimanSetPinBus(): Pin 1 successfully set to type GPIO (1) with bus 0x2
ESP32-DRV: Configuring 3PWM in group: 0 on timer: 0
ESP32-DRV: Configuring 2 operators.
ESP32-DRV: Configuring 3 comparators.
ESP32-DRV: Configuring 3 generators.
ESP32-DRV: Configuring center-aligned pwm.
ESP32-DRV: Enabling timer: 0
ESP32-DRV: MCPWM configured!
MOT: Monitor enabled!
MOT: Init
MOT: Enable driver.
Motor 0 ready.
[  1640][V][esp32-hal-periman.c:235] perimanSetBusDeinit(): Deinit function for type GPIO (1) successfully set to 0x42004b6c
[  1652][V][esp32-hal-periman.c:160] perimanSetPinBus(): Pin 10 successfully set to type GPIO (1) with bus 0xb
[...]

PR #442, opened 2024-11-07
ESP32 C6 fix

runger1101001:
Fix for ESP32 C6 as suggested here: https://github.com/simplefoc/Arduino-FOC/issues/440


runger1101001:
Also includes PR #430 and some other fixes.

PR #444, opened 2024-12-23
Add missing '\r' character sending to Commander

AntonEvmenenko:
Hello,

The default Arduino way to print a new line is to write two characters, `\r\n`. However, there is code in the Commander class that only uses the `\n` for the new line. In my case (definitely it depends on the OS and terminal program) it leads to incorrect behavior, something like this:

```
?
Warn: \r detected!
                   A: command A
B: command B
```

So I just replaced `\n` with `\r\n`

### DIFF
--- a/src/communication/Commander.cpp
+++ b/src/communication/Commander.cpp
@@ -592,7 +592,7 @@ bool Commander::isSentinel(char ch)
     return true;
   else if (ch == '\r')
   {
-      printVerbose(F("Warn: \\r detected! \n"));
+      printVerbose(F("Warn: \\r detected! \r\n"));
       return true; // lets still consider it to end the line...
   }
   return false;


runger1101001:
Sorry, but I think the \n line endings are generally preferred, and we won't be changing this.

There's no way to this right, since windows uses \r\n, Unix/Linux uses \n and Mac uses \r.

We prefer to use just \n for several reasons:

1) Its the way it is in Unix/Linux
2) The single character representation is shorter, which given that we have short messages saves a significant percentage of bits to transfer.
3) Since no one in the world is sending their SimpleFOC output to 1970s line printers, there is absolutely no value in sending both characters.

--> you can change the line ending character on the PC side:

Arduino: 

image

platformio.ini:

`monitor_eol = "\n"`

PR #447, opened 2025-01-17
Update esp32_ledc_mcu.cpp to set ledc_timer = false

rob-deutsch:


askuric:
Thanks for this!
This seems to be a API change actually going from the espressif 5.1 to 5.2
Introducing the deconfigure  might actually break the code for people with v3.0.x of the arduino-esp32 which depends on espressif 5.1.4. So we will have to handle this with care 😄

`ledc_timer_config_t` definition 
- in 5.1.4 - https://github.com/espressif/esp-idf/blob/a9424f8a356c1eb9f38cca506ff27cda6dc0eb89/components/driver/ledc/include/driver/ledc.h#L56
- in 5.2+ - https://github.com/espressif/esp-idf/blob/adf5319639b7e66e4f80868b14fee1acb73672c7/components/esp_driver_ledc/include/driver/ledc.h#L56

Could you just confirm which version of espressif and ardunio-esp32 are you using?

rob-deutsch:
Oh good pickup. I'm actually using pioarduino because espressif/arduino-esp32 doesn't seem to support platformio any more.

Maybe the simplest way forward is to calloc/memset ledc_timer_config_t to 0 based on its size?

rob-deutsch:
Any thoughts on why the tests are failing? I didn't change anything which should break those tests, and the error message isn't clear to me.

runger1101001:
> Any thoughts on why the tests are failing? I didn't change anything which should break those tests, and the error message isn't clear to me.

It was unrelated to your change and I have fixed it.

> Introducing the deconfigure might actually break the code for people with v3.0.x of the arduino-esp32 which depends on espressif 5.1.4. So we will have to handle this with care 😄

So how should we handle it? I must say I find the current situation with ESP32 very confusing. Some boards still work with the old framework versions, also in platformio. Other boards (C6, C5, P4) require the new framework. The new framework official release doesn't work in PlatformIO, but the new pioarduino framework does.

rob-deutsch:
> > Introducing the deconfigure might actually break the code for people with v3.0.x of the arduino-esp32 which depends on espressif 5.1.4. So we will have to handle this with care 😄
> 
> So how should we handle it?

Good question - the updated branch I forced push will work across any and all versions of arduino-esp32 and pioarduino. It's entirely cross platform because it just inits the struct to zeroes.

I would suggest merging this PR and tackling the question about ESP32 compatability elsewhere.

> I must say I find the current situation with ESP32 very confusing. Some boards still work with the old framework versions, also in platformio. Other boards (C6, C5, P4) require the new framework. The new framework official release doesn't work in PlatformIO, but the new pioarduino framework does.

I also find the current situation confusing. It's an unfortunate situation that seems to be caused by Espressif no longer supporting platformio in arduino-esp32.

I don't yet have a strong POV as to what the right thing to do is. I'm sure that simplefoc isn't the only one facing this problem.

runger1101001:
Ok, I'm still a bit confused by this line of the diff:

image

But I will merge that and solve the #endif separately if there is an issue with it. It looks like it was both removed and added, weird.

rob-deutsch:
I suspect that's an accidental addition of a linefeed (LF) character at the end of the line.

Would you like me to submit a PR which removes this additional LF character?

runger1101001:
No, thanks for offering this, but I have already merged it :-) I made the mistake of merging it to master, doh! That's what I actually should have checked, but never mind. Its really just a small change, and I'll fix all this in time for the next release...

rob-deutsch:
Sorry about that, I should've spotted it before submitting the PR. Thanks for your understanding.

PR #449, opened 2025-02-03
Current calculation typo in comments

askuric:
A wrong calculation of the current in the comments of the examples.


runger1101001:
Comment change only - can be safely merged...

PR #454, opened 2025-02-22
Add current and voltage feed forward terms to motor classes

Copper280z:
This change adds feed forward terms for the voltage and current controllers. This allows including inputs using system parameters and knowledge not available to the PID controllers to improve performance. 

Some examples of use cases might be using the moving inertia with an S-curve motion profile to calculate the torque (current) required to follow the profile. Without feed forward, following error depends solely on the PID controller, with feed forward the PID controller is only responsible for compensating unmodeled disturbances. Another example would be using a motor to resist a known load, like gravity, or a spring, while in angle or velocity control mode. This currently would depend on the PID controller to compensate for the load, but because the load is well known it can be calculated directly. The response of the system to predicable changes in this load no longer depends on the PID controller bandwidth.

This change still needs testing, but I wanted to get it out there while I had a chance. :)

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -355,23 +355,23 @@ void BLDCMotor::loopFOC() {
       // read overall current magnitude
       current.q = current_sense->getDCCurrent(electrical_angle);
       // filter the value values
-      current.q = LPF_current_q(current.q);
+      current.q = LPF_current_q(current.q) + feed_forward_current.q;
       // calculate the phase voltage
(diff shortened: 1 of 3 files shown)


runger1101001:
Hey @Copper280z thanks so much. This looks great, and I will merge it right after doing the 2.3.5 release. Its an important change, and we'll also have to update the docs accordingly, so I don't want to rush it for this release if that's ok?

Copper280z:
That’s totally fine by me, I’d rather have more testing time than less before it’s released.

PR #457, opened 2025-03-12
Feat hybrid stepper to main + implement current sensing for it

askuric:
Hi!

Ok so I've had some time to test the HybridStepperMotor class from the drivers repo and I love it.

It is a very nice addition to the library and could be a big one, enabling to run steppers with widely available BLDC drivers. 
Much easier than with LN298 and other old drivers. 

I've implemented the current sensing for it and verified that it works. It's still relative experimental but I think it could actually enter in the next release. Just because of it's potential.  :)

But it can also wait a bit of course. 

- This PR should be synced with: https://github.com/simplefoc/Arduino-FOC-drivers/pull/55

### DIFF
--- a/src/HybridStepperMotor.cpp
+++ b/src/HybridStepperMotor.cpp
@@ -1,0 +1,588 @@
+#include "HybridStepperMotor.h"
+#include "./communication/SimpleFOCDebug.h"
+
+// HybridStepperMotor(int pp)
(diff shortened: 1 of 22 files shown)


runger1101001:
This is great. It should be merged right away to make the stm32 current sensing work with the new PWM driver.

PR #459, opened 2025-03-14
use default dir only if sensor not connected

askuric:
Hey @runger1101001,

I think this is a bug, but I am not sure. Let me know if I'm missing something. 
I am not 100% sure what is the motivation behind this part of the code.

The issue that I was having is that if the user uses the openloop mode at the start the initFOC will skip the direction calibration (as the motor.init will set the direction to CW). And this can lead to unstable behaviour later if the real direction is CCW and the user goes to the closed-loop control. Does it make sense to set the default sensor direction only if the sensor is not there?

### DIFF
--- a/src/BLDCMotor.cpp
+++ b/src/BLDCMotor.cpp
@@ -93,10 +93,13 @@ int BLDCMotor::init() {
   P_angle.limit = velocity_limit;
 
   // if using open loop control, set a CW as the default direction if not already set
-  if ((controller==MotionControlType::angle_openloop
-     ||controller==MotionControlType::velocity_openloop)
-     && (sensor_direction == Direction::UNKNOWN)) {
-      sensor_direction = Direction::CW;
+  // only if no sensor is used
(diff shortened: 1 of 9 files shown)


runger1101001:
Hey Antun,

It would seem to me that after this change, and if the sensor is set, then in open loop mode the direction would remain in UNKNOWN (unless you explicitly set it)?
I think this is ok, as it isn't used in open loop mode. But if the user 

I think the motivation is that when you use the open loop mode in combination with a sensor. TBH I can't remember where it came from but I suspect it has to do with the python calibration & linearity measurement where I was running open loop but with the sensor in use.

I guess these and other similar examples are different edge cases - to handle them correctly its probably necessary for the users to explicitly set or unset the direction depending on what they want.

Its fine for me to remove this or change it in the way you have suggested, but I think there might be some of the users affected who are doing things like open loop + velocity measurement, or running their own calibration routines.

PR #460, opened 2025-03-19
Feat h7 current sensing

askuric:
- This PR adds the low-side current sensing for H7 boards. 
- There are some cleanups of the driver code as well, minimal though
- The current sensing does not work with portenta for the moment - did not have on me at the moment so I could not really test the MBED integration properly.
- Also I've added portenta and a few other MBED boards to the compile CI


askuric:
H7 low-side is working, I've tested it with the SimpleFOCMini v2. 
Before this [commit](https://github.com/simplefoc/Arduino-FOC/pull/460/commits/302719c14102e286158472d4de04b98f3efabc97) it did not work properly actually. It did not handle the repetition counter missing. 

I did find something strange though. In the case where the repetition counter is handled in the software (in the interrupt handler) where we count and use every second measurement, for H7 we have to start from the 2nd call of the interrupt while the other architectures were starting from the 1st call. There is a shift of one interrupt call. I am not sure where does it come from. 
I'm wondering if this shift is persistent between the architectures (always the same for the same family) or is it different for different timers, or is there something else going on. In any case I did not see this before. 

We need to investigate this!

runger1101001:
How about this one? It still has conflicts...

askuric:
Should be all good now!

runger1101001:
Lets merge it :-)

PR #463, opened 2025-04-19
Use AS5600 auto increment suppress

tikonen:
This IC does not increment I2C read pointer when the angle is read for performance reasons. There is no need to rewrite the register address on an every request.

This makes angle query practically almost twice as fast.  I've tested this only on AS5600 so feature is disabled by default for other I2C comms based sensors.

### DIFF
--- a/src/sensors/MagneticSensorI2C.cpp
+++ b/src/sensors/MagneticSensorI2C.cpp
@@ -5,7 +5,8 @@ MagneticSensorI2CConfig_s AS5600_I2C = {
   .chip_address = 0x36,
   .bit_resolution = 12,
   .angle_register = 0x0C,
-  .data_start_bit = 11
+  .data_start_bit = 11,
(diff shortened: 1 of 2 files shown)


runger1101001:
Hi @tikonen , thank you for contributing this but actually we already support this in our dedicated driver for the AS5600:

https://github.com/simplefoc/Arduino-FOC-drivers/tree/master/src/encoders/as5600

This driver additionally keeps the I2C transactions open and in my tests achieves significantly higher read rates than our regular driver.
It also ha stehe advantage of being able to read/write the sensors other registers.
So I think I’d prefer to keep the MagneticSensorI2C driver really basic and free of hardware-specific functions, since we have the dedicated drivers for the people who want them.

tikonen:
No problem. My thinking was as the driver had explicit configuration for AS5600 this might go in as optional functionality but it makes sense to keep it simple if there is already dedicated high performance driver available.

PR #468, opened 2025-05-07
Add support for timer 8

jeroenzwan:
We use timer 6,7,8 for phase a,b,c. Timer 8 is only available in RP2350B, not 2040 or A. Therefore, the library does not support it well. We found that everything works except for the syncSlices function which sets pwm mask enabled with a bitmask 0xFF, so timer 0 to 7. Timer 8 was never enabled and did not run.


runger1101001:
Thank you so much for finding and solving this!

While I did a quick test on the Pico 2, I did not go into this depth to test the new timer... I'll merge it in as soon as the tests pass.

PR #469, opened 2025-05-08
Fix hall sensor velocity estimation

ahughesattomra:
The hall sensor velocity estimation could end up returning a nonzero value if the timer "wrapped" between runs, even if the motor was actually stationary. If using the velocity regulator the motor may then fail to start.

To solve this:
- Switch to using unsigned types so that wrapping is defined.
- When timing out the pulse value, also reset it so that the velocity won't be briefly treated as valid again when the timer wraps.

Seen on a platform with 32 bit long, so running the motor and then waiting ~40 minutes is enough to observe this bug. Note that the pulse_diff variable is reset if the hall sensor direction changes, which may happen naturally when the motor stops, depending on what it is attached to.

### DIFF
--- a/src/sensors/HallSensor.cpp
+++ b/src/sensors/HallSensor.cpp
@@ -122,10 +122,19 @@ float HallSensor::getSensorAngle() {
 */
 float HallSensor::getVelocity(){
   noInterrupts();
-  long last_pulse_timestamp = pulse_timestamp;
-  long last_pulse_diff = pulse_diff;
+  unsigned long last_pulse_timestamp = pulse_timestamp;
+  unsigned long last_pulse_diff = pulse_diff;
   interrupts();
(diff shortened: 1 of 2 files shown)


ahughesattomra:
Only tested on v2.3.1 on a non-Arduino platform.

The bug is real though, so it'd be nice if it was fixed upstream.

runger1101001:
Thank you very much for contributing this. I can't easily test it at the moment, but we'll review it and then include it in an upcoming release.

dekutree64:
Good catch! I think the integer wrapping behavior will work just as well with signed or unsigned, but setting pulse_diff = 0 is indeed necessary to prevent returning the previous valid velocity reading if it sits still for half a _micros() wrap. I think it would even continue to return that reading for the next 36 minutes until _micros() fully wraps back to where it started.

PR #472, opened 2025-05-20
Fix stepper motor low side current sensing

AntonEvmenenko:
See the discussion in Discord starting from [here](https://discord.com/channels/927331369137352734/927333189981196298/1374189697391132712)


runger1101001:
Thanks a lot for contributing this! I know you're in active discussions with Antun, who is working on the Stepper current sensing, so I'll leave this to him...

PR #474, opened 2025-06-17
Disable allow power down when creating MCPWM timer

adamgerhant:
The ESP driver code returns an error if the chip does not support retention of registers during power down when creating a MCPWM timer.

[Driver code](https://github.com/espressif/esp-idf/blob/adcbdd7da4fca12cd48ca5338e3fa0e8cf6695ce/components/esp_driver_mcpwm/src/mcpwm_timer.c#L85):
```
#if !SOC_MCPWM_SUPPORT_SLEEP_RETENTION
    ESP_RETURN_ON_FALSE(config->flags.allow_pd == 0, ESP_ERR_NOT_SUPPORTED, TAG, "register back up is not supported");
```

Also adds MT6701 to magnetic sensor i2c export


runger1101001:
Thank you very much for contributing this!

We only accept PRs to the dev branch, so I changed the base of your PR.

The ESP32-C3 test is failing, but I think its nothing to do with your PR.
The declaration of the MT6701 SPI settings was in fact already present on the dev branch. I've resolved the conflict in the PR.

I will merge the fix for the register retention. Thanks a lot for finding this!

PR #476, opened 2025-07-08
fixed MT6701 I2C

charizardavi:
updated MagneticSensorI2C.h to include MT6701_I2C. Previously threw compile error due to missing declaration.


runger1101001:
Thank you very much for reporting and contributing this!

However, this issue is already fixed on the dev branch, and will be part of the next release...
see https://github.com/simplefoc/Arduino-FOC/blob/e235d706c5917c0e4ea6bfec1537c2678eb1ba16/src/sensors/MagneticSensorI2C.h#L21

PR #487, opened 2025-09-25
g431-esc: use internal ADC channels, also fix some typos

Moddingear:
Before this commit, the opamps amplify, then the signal goes outside to the pin, where it is taken back into the microcontroller to be converted. 

The argument for this was that it allows filtering, but since the pin goes in and out, that would rely entirely on the output impedance of the opamp to do filtering, and in the case of this board, only one pin is actually broken out as a test pad.

This changes that and makes the board use the internal opamp to ADC channels. It is also now possible to use -1 to -3 to access the opamp readings, since in other boards, these may not have any pins assigned.

This change can be turned off by a define.


runger1101001:
Ok so let’s merge it? What’s the status of your testing it? I’m hesitant because I don’t have a test setup for it myself.

PR #488, opened 2025-09-29
Fix issue with angle wrap-around

rengels:
The 0.8 factor was sub-optimal.

Let's assume an angular velocity of 0.3*2PI per calls of update.
Then a wrap-around will produce a change of 2PI - 0.3*2PI, that's 0.7*2PI
So it won't be detected as a wrap-around of the angle at all and
thus the number of full rotations will be off, as will be the
velocity.

The perfect number is 0.5.

### DIFF
--- a/src/common/base_classes/Sensor.cpp
+++ b/src/common/base_classes/Sensor.cpp
@@ -6,12 +6,22 @@
 
 void Sensor::update() {
     float val = getSensorAngle();
-    if (val<0) // sensor angles are strictly non-negative. Negative values are used to signal errors.
+
+    // sensor angles are strictly non-negative. Negative values are used to signal errors.
+    if (val < 0) {
         return; // TODO signal error, e.g. via a flag and counter
+    }
+
     angle_prev_ts = _micros();
     float d_angle = val - angle_prev;
-    // if overflow happened track it as full rotation
-    if(abs(d_angle) > (0.8f*_2PI) ) full_rotations += ( d_angle > 0 ) ? -1 : 1; 
+
+    // if wrap-around happened. track full rotation
+    if (d_angle > (0.5f *_2PI)) {
+        full_rotations--;
+    } else if (d_angle < -(0.5f *_2PI)) {
+        full_rotations++;
+    }
+
     angle_prev = val;
 }
 
--- a/src/common/base_classes/Sensor.h
+++ b/src/common/base_classes/Sensor.h
@@ -109,15 +109,14 @@ class Sensor{
         float min_elapsed_time = 0.000100; // default is 100 microseconds, or 10kHz
 
     protected:
-        /** 
-         * Get current shaft angle from the sensor hardware, and 
+        /** Get current shaft angle from the sensor hardware, and
          * return it as a float in radians, in the range 0 to 2PI.
-         * 
+         *
          * This method is pure virtual and must be implemented in subclasses.
          * Calling this method directly does not update the base-class internal fields.
          * Use update() when calling from outside code.
          */
-        virtual float getSensorAngle()=0;
+        virtual float getSensorAngle() = 0;
         /**
          * Call Sensor::init() from your sensor subclass's init method if you want smoother startup
          * The base class init() method calls getSensorAngle() several times to initialize the internal fields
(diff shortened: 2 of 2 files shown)


runger1101001:
This has been something I have noticed for a while, and has been bothering me also.

I agree with your conclusion that the best check would be for 0.5 x 2PI, or just PI radians of difference.
But I don’t agree with your maths: 0.32rad/update would result in a difference of 1.68xPI, which is > 2PI x 0.8, so would still be detected as a wrap-around.

I also don’t think it’s a real-world problem, TBH, since following your example, assuming we have a very slow update frequency of only 1kHz, then 0.32rad/update would mean a real world speed of 320rad/s, > 3000RPM. But going 3000RPM with only 1kHz update rate is a very unreasonable expectation, it won’t happen IRL…

So to reach such speeds you’d expect maybe 10kHz or at least 5kHz update rate, in which case you’ll be far from the 0.32rad/update again…

So I think this topic is only a theoretical problem, not a real one. If you’re coming close to missing the wrap-around, then it’s your iteration speed that needs to improve, not the 0.8 bs 0.5 factor….

src/common/base_classes/Sensor.cpp
```diff
@@ -6,12 +6,22 @@
...
     float d_angle = val - angle_prev;
-    // if overflow happened track it as full rotation
-    if(abs(d_angle) > (0.8f*_2PI) ) full_rotations += ( d_angle > 0 ) ? -1 : 1; 
+
+    // if wrap-around happened. track full rotation
+    if (d_angle > (0.5f *_2PI)) {
+        full_rotations--;
+    } else if (d_angle < -(0.5f *_2PI)) {
+        full_rotations++;
+    }
```
austin-bowen, line 23: I recommend just using the `_PI` constant to cut down on math ops

austin-bowen:
I agree with making this improvement; anecdotally, I've found it improves the rate of successful full rotation detections for my use case, which is low-RPM (60-120 ish).

In my case, I'm using an analog position sensor which outputs a scalar voltage from 0..Vcc over the full rotation range, then wraps back around to 0V. I've been running into problems where the code isn't always recognizing when a full rotation has occurred.

It's hard to prove, but I suspect it's because SimpleFOC is sampling faster than the sensor can drive the output voltage from Vcc to 0, so SimpleFOC is sampling "on the way down" and reading the intermediate voltages as actual angles, resulting in occasional angle diff sequences like (made up worst-case example):

```
+0.1 deg
+0.1 deg
+0.1 deg
-180 deg  <-- wrap-around begins
-180 deg
+0.1 deg
+0.1 deg
...
```

which does not pass the 0.8*2pi threshold.

Changing the threshold to just pi doesn't completely solve this particular problem, but it helps.

src/common/base_classes/Sensor.cpp
rengels: The "math ops" is compile time, but it makes sense.

austin-bowen: I don't think it's guaranteed the `0.5f * _2PI` will be optimized to a single constant by the compiler

rengels:
The "scalar voltage" angle sensor would possibly need some additional logic. 
You will always get intermediate voltages depending on the sample speed and the strength of the angle sensor output driver.

Actually you will always get intermediate voltages.
There will be a sampling happening at the wrong time once in a while.
I could imagine that my patch helps in this case.

austin-bowen:
> Actually you will always get intermediate voltages.
> There will be a sampling happening at the wrong time once in a while.

Yes exactly -- I didn't mention it, but I am working on a "fix" of sorts. Basically just capping the sample rate of raw sensor readings to e.g. 1kHz, like `Sensor::getVelocity()` does for velocity measurements. Not bullet-proof, but reduces the odds.

rengels:
> > Actually you will always get intermediate voltages.
> > There will be a sampling happening at the wrong time once in a while.
> 
> Yes exactly -- I didn't mention it, but I am working on a "fix" of sorts. Basically just capping the sample rate of raw sensor readings to e.g. 1kHz, like `Sensor::getVelocity()` does for velocity measurements. Not bullet-proof, but reduces the odds.

It will only reduce the odds.
Let's say you make one sample per second and your signal flank in case of a wrap-around is .1ms
Then you still have the problem once every three hours.

With my patch it will look like this:

deg | simple FOC angle | comment
------|----------|----
+0.1 deg | 359.7 | -
+0.1 deg | 359.8 | -
+0.1 deg | 359.9 | -
-179 deg | 181 | - <-- wrap-around begins, but one meassurement is below 180deg and one above
-181.1 deg | 360.1 | -
+0.1 deg | 360.2 | -
+0.1 deg | 360.3 | -

Still ugly, but at least it doesn't miss a complete revolution

PR #493, opened 2025-10-15
ESP32: replaced the mod base circular increment

uLipe:
The current low side sense of the ESP32 uses a mod based circular buffer increment which may consume some precious cycles, we address this issue by replacing it with the classic branch based circular increment which outspeed  the current solution in more than 5 times.


uLipe:
@runger1101001 @askuric PTAL :)

runger1101001:
I assume the improvement is related to just this part of the code, where avoiding the division implied by modulo operator saves a lot compared to a simple compare operation. It might be even more robust to make it a >= comparison though...

uLipe:
@runger1101001  just changed the comparison, also I re-ordered the ADC buffer index increment to happen after the ADC is read it was affecting the order of the phase-currents since the phase a was becoming swapped with second (on 2-phase measurement) and with the c (in 3 phase measurement setup).

PR #494, opened 2025-10-23
FOCDriver: Replace the divisions with reciprocal multiplication.

uLipe:
The SetPWM functions of all the drivers perform a constrain operation that normalizes the phase voltage and DC link voltage using a division operation, which may take several cycles even on processors that feature a FPU on their hardware.

This waste of cycles (which represent a precious amount of microseconds) can be drastically reduced when replacing the division by reciprocal multiplication, since the power supply voltage value will not change in the runtime it it is valid to replace for example:

```
 dc_a = _constrain(Ua / voltage_power_supply, 0.0f , 1.0f );
```

To :

```
 dc_a = _constrain(Ua * inverse_power_supply, 0.0f , 1.0f );
```

where:

```
inverse_power_supply = 1.0f / voltage_power_supply;
```
The `inverse_power_supply` is calculated during the driver initialization avoiding the division to be done in the SetPWM function which is a hot code path, that is it, it gets called very often in the FoC main loop. I took the care to make this generic by introducing this new member in the base class and doing the inverse calculation in all the drivers available.

Probably this optimization can be reused in more hot code paths, I will check where this can be potentially applied.

### DIFF
--- a/src/common/base_classes/FOCDriver.h
+++ b/src/common/base_classes/FOCDriver.h
@@ -35,6 +35,7 @@ class FOCDriver{
         long pwm_frequency; //!< pwm frequency value in hertz
         float voltage_power_supply; //!< power supply voltage
         float voltage_limit; //!< limiting voltage set to the motor
+        float inverse_power_supply; //!< inverse of the power supply, very useful to replace divisions with reciprocal multiply which is faster
 
         bool initialized = false; //!< true if driver was successfully initialized
         void* params = 0; //!< pointer to hardware specific parameters of driver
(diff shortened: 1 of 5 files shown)


runger1101001:
Yeah, it’s true, but the fix has to be more complicated unfortunately.

this is because the assumption that voltage_power_supply remains constant after initialization is not correct.

A more complete code for a ESC will include VBUS voltage monitoring and adjust the power supply voltage to match the real world conditions.

It’s also unfortunate that the current code performs multiple scalings of the voltage, because the motor class deals with voltages, while the driver.setPwm() converts to 0-1 duty cycles and the hardware drivers deal in integer counter values.
I would prefer to see a solution where we skip the intermediate 0-1 representation and go directly to integer values in timer counts.

PR #503, opened 2025-12-23
initial support for low-side current sensing with multiple motors for stm32

askuric:
Ok so this is the bare minimum support for having multiple motors run at stm32 boards with low-side current sensing. 

- Each motor has to go on their own ADC!
- There are no any other additional checks than the one ADC per motor
- The code should be completely backward compatible (no change in the API)


runger1101001:
I think this is a big improvement for people wanting to run multiple motors.

There are still a lot of complexities and sub-cases to consider, but at least this makes it possible to do it at all... from my point of view I would merge it...

PR #504, opened 2025-12-23
Autotunng + a big reactor without API chnage

askuric:
So this is a PR that does two things:

1. Does a big refactor of the FOCMotor code with the aim to simplify the BLDC, Stepper and HybridStepper motor codes
2. Implements a more flexible autotuning from the #426 

## Refactor
- All the common functionalities moved to the FOCMotor class 
   - Sensor alignement
   - Current sese alignment
   - Motion control
   - Torque control
   - Calibration ... 
- Only the truly different parts left in the BLDCMotor, StepperMotor, ... 
   - Setting the phase voltages 
   - init,enable,disable
   
## Autotuning
- Baiscally the code is more or less the same as in #426 
- But it is more flexible and allows to be called from the Commander - ex `MFC100` for the bandwith of `100`Hz
- It can be also called in the `setup`. - ex `motor.tuneCurrentController(100)`
- It will either use the motor parameters 
   - from the constructor or
   - if the current sense is available and no params are provided it will calll `motor.characteriseMotor`

### DIFF
--- a/src/common/base_classes/FOCMotor.h
+++ b/src/common/base_classes/FOCMotor.h
@@ -46,8 +46,9 @@ enum MotionControlType : uint8_t {
   torque            = 0x00,     //!< Torque control
   velocity          = 0x01,     //!< Velocity motion control
   angle             = 0x02,     //!< Position/angle motion control
-  velocity_openloop = 0x03,     //!< Open loop velocity control
-  angle_openloop    = 0x04      //!< Open loop position/angle control
+  velocity_openloop = 0x03,
+  angle_openloop    = 0x04,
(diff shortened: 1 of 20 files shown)


runger1101001:
This seems like a good refactoring to me...

One thing to maybe talk about is to make the PID and LPF classes exchangeable instead of hard-coded, and then instead of combining fixed and variable sample times into one class, to make separate classes and therefore also open the possibility to have alternative implementations.

askuric:
I like the idea. Do you have an idea on how to do it though?
linkLPF mothod or something like that?

PR #506, opened 2026-01-07
Regular channel read for stm32 (while injected ADC conversion runnig)

askuric:
# Description

This PR is an initial proposition of the reagular ADC channel reads on stm32 that 
- Is more efficient than `analogRead`
- Allows reading regular ADC channels even if the Injected channles are read (ex. low-side current sesnig)

**NOTE**
It still uses the polling strategy, no DMA so it is still blocking and is not intended to be called to often. 
On f405 boards the ADC read time is around 20us and on g4 is similar. 
For reference analogRead takes about 60us+

**Tested setups**
Tested on f4 and g4 boards with different setups. 
Did not test on other boards though, but none of the original API calls shouldn't be affected by this.


askuric:
Works on H7 as well - Nucleo h723zg 
`_readRegularADCVoltage` takes about 3us

PR #523, opened 2026-04-09
Fix potential undefined behavior in sin/cos

### LINKED ISSUE #505: [BUG] Potential Undefined Behavior in _sin/_cos due to float-to-unsigned cast
**Issue Description:**
I have identified a potential logic bug in the optimized trigonometric functions in `foc_utils.cpp`. The code currently performs a direct cast from a `float` to an `unsigned int`, which results in **undefined behavior** when the input angle `a` is negative.

**Affected Code:**
In [`src/common/foc_utils.cpp`](https://github.com/simplefoc/Arduino-FOC/blob/master/src/common/foc_utils.cpp):

* **`_sin(float a)`**
* **`_cos(float a)`** (indirectly)

```cpp
// Current implementation
unsigned int i = (unsigned int)(a * (64*4*256.0f/_2PI));
```

**Technical Explanation:**
According to the C++ standard (ISO/IEC 14882), casting a negative floating-point value to an unsigned integer type is undefined.

While the library typically calls `_normalizeAngle` before using these functions in the main FOC loop, if a user or a different part of the library calls `_sin()` with a negative angle, the `unsigned int` cast can result in a value that breaks the quadrant logic.

For example, on many MCUs, a negative float cast to `unsigned int` will saturate to `0` or wrap to a massive integer, leading to an incorrect index `i` and an incorrect sine result.

**Suggested Fix:**
Changing the type of `i` from `unsigned int` to a signed `int` ensures the cast is well-defined. The subsequent bitwise operations (`& 0xff` and `>> 8`) will still function correctly to resolve the table index and fractional part.

```cpp
// Suggested change
int i = (int)(a * (64*4*256.0f/_2PI));
```

dekutree64:
# Description

Fixes #505 (cast from float to unsigned int not guaranteed to work on all platforms)

When testing, I also discovered that it may not work on platforms with 16-bit int. On PC, casting from float to short (signed 16-bit integer) will saturate to -32768 if the float value is outside of the -32768 to 32767 range, rather than truncating the upper bits. So the initial conversion is now to 32-bit int, then truncate to whatever is fastest for the platform you're on.

## Type of change

# How Has This Been Tested?

Max error and RMS error verified unchanged on PC. Tested running a motor in open loop mode on STM32.


dekutree64:
Forgot to change branch to dev

PR #525, opened 2026-04-12
MagneticSensorAnalog: Correct angle calculation by subtracting min_raw_count from raw_count

austin-bowen:
# Description

`MagneticSensorAnalog::getSensorAngle` has a slightly incorrect implementation for converting `raw_count` to degrees:

> `return ( (float) (raw_count) / (float)cpr) * _2PI;`

It should first subtract `min_raw_count` from `raw_count` so it is properly in the range `[0, cpr]` before dividing by `cpr`.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
^ Not sure which of the two it is. I'd expect it to only make things better due to more accurate position calculations, but hard to be sure somebody's current code doesn't somehow depend on the slightly incorrect original implementation.

# How Has This Been Tested?

Ran velocity control, which was working before my change, and worked after as well.

**Test Configuration/Setup**:
* Hardware:
  * SimpleFOC Mini
  * 2804 gimbal motor
  * AS5600 magnetic encoder (using analog OUT)
  * All from [this Amazon listing](https://www.amazon.com/dp/B0FXKN9YMJ)
* IDE: Arduino
* MCU package version: XIAO RP2040


austin-bowen:
smh I didn't fork from `dev`; closing, will re-open

austin-bowen:
New PR: https://github.com/simplefoc/Arduino-FOC/pull/526

Issue #302, opened 2023-09-02
[FEATURE] An idear for mechanic_angle 


FFiot:
1. mechanic_angle & mechanic_angle_prev from 0° ~ 360°.
2. The angle changes from -180° to +180°.
3. rotation_cnt for angle control.
 
step 1:
if(mechanic_angle > mechanic_angle_prev + 180){
    mechanic_angle -= 360;
}else if(mechanic_angle < mechanic_angle_prev - 180){
    mechanic_angle -= 360;
}

step 2:
LPF_angle(mechanic_angle);

step 3:
if(LPF_angle.y_prev >= 360){
    LPF_angle.y_prev -=360;
    rotation_cnt += 1; 
}
if(LPF_angle.y_prev < 0){
    LPF_angle.y_prev +=360;
    rotation_cnt -= 1；
}

runger1101001:
I am not sure I understand… what problem are you trying to improve with this?

the shaft angle is already computed via LPF filtering, and the full rotations are accounted for in the sensor class…

FFiot:
float Sensor::getAngle(){
    return (float)full_rotations * _2PI + angle_prev;
}
Do not need：(float)full_rotations * _2PI 。

runger1101001:
I think internally it already works this way - see sensor.getMechanicalAngle()

the precision problem exists still in position mode - where you can command angles mich larger than 2PI

we would have to change the API for position mode to solve this problem.

on the sensor side it is already solved…

dekutree64:
I do think this is the proper way to handle large angles, but it has a fairly large performance penalty to solve a problem that only affects a few users. The PID code in motor.move will have more added cost, and motor.target is still trouble. 

I'm more inclined to abandon the idea of handling large angles entirely, and instead provide a way for users to manually wrap the angle back into the safe range when needed.
```
int32_t FOCMotor::wrapAngle(){
  int32_t full_rotations = sensor->wrapAngle();
  shaft_angle -= full_rotations*_2PI;
  LFP_angle.y_prev -= full_rotations*_2PI;
  return full_rotations;
}
```
LowPassFilter::y_prev will need to be made public or FOCMotor added as a friend class, and several sensor classes will need to implement wrapAngle, which in the case of the base class will simply be `int32_t Sensor::wrapAngle(){int32_t temp = full_rotations; full_rotations = 0; return temp;}`

[...]

PR #530, opened 2026-04-25
Improve _normalizeAngle speed.

jaguilar:
# Description

Previously _normalizeAngle used fmodf, which is expensive on Cortex-M series chips (40-100 cycles). For normalizing the angle, we can use a trick of just subtracking 2 pi times the number of full rotations from the angle.

This method does lose precision much faster than the fmodf approach. There will be significant error if the total number of radians passed in is greater than 10^5. However, we can counteract this by not passing in such a large number of radians -- basically making sure that we are not using an accumulating radian count in any location where the angle will eventually need to be normalized. This should typically be the case since accumulating radian use cases are definitionally not in need of normalization. We should consider adding an assert here to ensure that we catch any existing cases where we should be passing the mechanical or electrical angle and are instead passing an accumulator.

Performance test (on STM32G474, with certain other optimizations):

Before:   foc_nanos:13737
After:    foc_nanos:12550 (approx 9% reduction in loopFOC time)

## Type of change

- [ ] Optimization

# How Has This Been Tested?

I built an arduino-foc-based test program based on this commit and observed that the motor still worked fine. I also benchmarked the code and observed it to be faster. I have not tested all possible control modes so it is not impossible that this approach is unacceptable in some configurations. Please review carefully and suggest anything further that I should check.

**Test Configuration/Setup**:
* Hardware: STM32G474 with custom gate driver board
* IDE: STM32CubeMX/VSCode/CMake
* MCU package version (stm32duino/arduino-esp32/..): N/A, using my own custom Arduino shims

### DIFF
--- a/src/common/base_classes/FOCMotor.cpp
+++ b/src/common/base_classes/FOCMotor.cpp
@@ -1,5 +1,6 @@
 #include "FOCMotor.h"
 #include "../../communication/SimpleFOCDebug.h"
+#include "common/foc_utils.h"
 
 /**
  * Default constructor - setting all variabels to default values
@@ -431,6 +432,9 @@ float FOCMotor::angleOpenloop(float target_angle){
     shaft_angle = target_angle;
     shaft_velocity = 0;
   }
+  // Normalize the shaft angle after each iteration to prevent it growing indefinitely
+  // and eventually losing precision.
+  shaft_angle = _normalizeAngle(shaft_angle);
 
   // save timestamp for next call
   open_loop_timestamp = now_us;
@@ -864,7 +868,7 @@ int FOCMotor::alignSensor() {
     // move one electrical revolution forward
     for (int i = 0; i <=500; i++ ) {
       float angle = _3PI_2 + _2PI * i / 500.0f;
-      setPhaseVoltage(voltage_align, 0,  angle);
+      setPhaseVoltage(voltage_align, 0, _normalizeAngle(angle));
 	    sensor->update();
       _delay(2);
     }
@@ -874,7 +878,7 @@ int FOCMotor::alignSensor() {
     // move one electrical revolution backwards
     for (int i = 500; i >=0; i-- ) {
       float angle = _3PI_2 + _2PI * i / 500.0f ;
-      setPhaseVoltage(voltage_align, 0,  angle);
+      setPhaseVoltage(voltage_align, 0, _normalizeAngle(angle));
 	    sensor->update();
       _delay(2);
     }
@@ -914,7 +918,7 @@ int FOCMotor::alignSensor() {
     sensor->update();
     // get the current zero electric angle
     zero_electric_angle = 0;
-    zero_electric_angle = electricalAngle();
+      zero_electric_angle = _normalizeAngle(electricalAngle());
     _delay(20);
     SIMPLEFOC_MOTOR_DEBUG("Zero elec. angle: ", zero_electric_angle);
     // stop everything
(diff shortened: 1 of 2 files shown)


src/common/base_classes/FOCMotor.cpp
```diff
@@ -440,6 +432,9 @@ float FOCMotor::angleOpenloop(float target_angle){
     shaft_angle = target_angle;
     shaft_velocity = 0;
   }
+  // Normalize the shaft angle after each iteration to prevent it growing indefinitely
+  // and eventually losing precision.
+  shaft_angle = _normalizeAngle(shaft_angle);
```
runger1101001, line 437: Did you test this? It seems to me we’ll never reach the target as shaft-angle is used to track the current position in open loop angle mode?

jaguilar: This part was not tested with angles greater than 2PI.  I agree that if the user sets angles outside the range, it wouldn't work. We should probably also normalize the target angle beforehand. WDYT?

jaguilar:
Okay, here is my suggestion. Let's drop all the parts of this commit that call _normalizeAngle in more places. Currently the only real place where _normalizeAngle is called on an accumulator is the shaft_angle of angle_openloop. I don't think it's very important to get high precision for very high angles here. If someone is willing to set a target angle above 50k, They will potentially lose some precision, but I don't think that is a typical use case, especially not in openloop mode.

My upcoming changes to _sincos will require normalizing some things that aren't currently normalized, but we will do that closer to the callsite. wdyt?

dekutree64:
Agreed. Changing to floorf is good, but eliminating the need to normalize electrical angle was one of the big wins of the current _sincos over the original non-interpolated 200-entry quarter-wave LUT. I'm curious to see what you have in mind if the added cost of normalizing would still come out ahead. If you're using the STM32's CORDIC, be aware that we already have some code for it here https://github.com/simplefoc/Arduino-FOC-drivers/tree/master/src/utilities/stm32math

For dealing with large angle precision, I still stand by my proposal here https://github.com/simplefoc/Arduino-FOC/issues/302

jaguilar:
> If you're using the STM32's CORDIC, be aware that we already have some code for it here https://github.com/simplefoc/Arduino-FOC-drivers/tree/master/src/utilities/stm32math
> 
> For dealing with large angle precision, I still stand by my proposal here #302

Yes, that's exactly the problem. That code uses fmod which is very very slow.

dekutree64:
My recommendation would be to use `CORDIC->WDATA = (q31_t)(a * (32768.0f / _PI)) << 16;`, which should give equal resolution and overflow range as the interpolated LUT _sincos, and effectively one CPU cycle to do the normalization.

But IIRC the CORDIC sincos with fmod was roughly the same speed as LUT before, so even if you do go with the floor trick it should be a nice speedup plus precision boost compared to the LUT. But only on STM32, so I think we should keep the normalization inside the CORDIC sincos rather than reintroducing it to the motor code.

For other places in the code where normalization is necessary, the floor trick is an excellent improvement. I never knew it was so much faster than fmod.

It's too bad we can't just use 16-bit fixed-point angles everywhere. The tricks you can do are so much fun. 32-bit for large angle values, cast to u16 for 0-2pi like _normalizeAngle, or cast to s16 for -pi,+pi representation. And sine/cosine are super fast without having to scale the range and cast to int.

jaguilar:
If we're good with the precision-reduced cordic code, that definitely reduces the need to normalize. That would be great.

PR #531, opened 2026-04-27
Replace double literals and fns with floats.

jaguilar:
# Description

In all cases, these literals are intended to be used in floating point operations. Floating point acceleration is much more common on devices where SimpleFOC runs. Even where no acceleration is present, soft floats are universally more efficient than soft doubles.

Additionally, fix several cases where fabs or flog are used instead of fabsf or flogf.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Ran my SimpleFOC test program.

**Test Configuration/Setup**:
* Hardware: STM32G464 Nucleo board
* IDE: Plain STM32 with Arduino wrapper
* MCU package version (stm32duino/arduino-esp32/..): N/A


runger1101001:
Very nice.

PR #536, opened 2026-05-17
fix: HybridStepper: center around supply voltage

schnili:
# Description

The Hybrid stepper centered its voltage around the driver voltage limit and not the current supply voltage. In the case the voltage limit of the driver is bigger than the supply this leads the center not really being the center.

## Type of change

To respect the current logic the center point is now `min(supply_voltage, driver_max_voltage)`

# How Has This Been Tested?

* Hardware: esp32s3 with custom DRV8323 driver (with `voltage_power_supply` measurement)
* IDE: Platformio
* MCU package version: arduino-esp32

### DIFF
--- a/src/HybridStepperMotor.cpp
+++ b/src/HybridStepperMotor.cpp
@@ -134,7 +134,7 @@ float HybridStepperMotor::estimateBEMF(float vel){
 // Function implementing Sine PWM and SVPWM algorithms
 void HybridStepperMotor::setPhaseVoltage(float Uq, float Ud, float angle_el)
 {
-  float center;
+  const float center = fminf(driver->voltage_power_supply, driver->voltage_limit) / 2;
   float _sa, _ca;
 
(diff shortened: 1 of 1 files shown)


runger1101001:
This seems to make sense from a code point of view, but from a use-case point of view, what would be the point of setting the voltage limit higher than the power supply voltage?

It seems to me this is checking for what is essentially a configuration error, but at the level of the setPwm(). 

I'd prefer to check it in the `updateVoltageLimit(float new_voltage_limit)` function, and mention it in the documentation (which it is), and leave it there rather than incur the runtime penalty of checking in each call to setPwm()...

schnili:
hm, then i have to check the doc again what the driver voltage is supposed to do.

This did come up as a problem as i am updating voltage_power supply to some thing that gets measured. if then it is e.g. going from 24 to 20 gradualy it goes below the driver limit.

PR #543, opened 2026-06-02
feat(esp32): digital ADC + DMA for MCPWM low-side current sense

uLipe:
## Summary

Adds **digital SAR ADC + DMA** for ESP32 **MCPWM low-side** current sensing (ported from espFoC `isensor_adc`), integrated into existing `LowsideCurrentSense` without API changes.

Three backends, selected at init:

| Mode | Targets | PWM → ADC | Sampling |
|------|---------|-----------|----------|
| **ADC_READ** (default) | All MCPWM ESP32 | MCPWM comparator/timer ISR | `adcRead()`, one phase per ISR (~10 µs each) |
| **DIGI_SW** | ESP32, S2, … | MCPWM ISR | `esp32_adc_digi_trigger_software()` — digi pattern + DMA |
| **DIGI_ETM** | S3, C6, … | **ETM** (TEZ → `ADC_TASK_START0`) | Same digi + DMA, no CPU trigger |

DMA shims: I2S (ESP32), SPI3 (S2), GDMA (S3/C6+). Pattern rate **80 kHz**, **2 channels**, `convert_limit = 2` (aligned with espFoC).

## Why this approach

Low-side FOC needs currents at a **fixed PWM phase**. The default **ADC_READ** path uses MCPWM callbacks + `adcRead()`, which:

- **Blocks the ISR** during each SAR conversion
- **Round-robins** phases across interrupts (latency/jitter between `i_a` / `i_b`)
- Limits headroom at higher PWM frequencies

**DIGI_*** modes run the hardware pattern sequencer and move results by DMA. On S3/C6, **ETM** starts the ADC without a CPU ISR on the trigger edge.

## Zero-copy DMA path

For `DIGI_SW` / `DIGI_ETM`, DMA writes `adc_digi_output_data_t[]` **directly into** `ESP32CurrentSenseParams::adc_buffer` (DMA-capable `heap_caps` alloc). The EOF ISR only:

- invalidates cache (where required), and  
- re-arms for the next conversion (ETM mode),

with **no per-sample copy loop**. `_readADCVoltageLowSide()` decodes the 12-bit field from the digi format on access.

## Problems addressed

- Shorter MCPWM ISRs on digi paths (no `adcRead` busy-wait in the trigger ISR on ETM; SW trigger only kicks the controller)  
- Both shunt channels in **one** DMA frame (2-shunt setups)  
- Lower trigger jitter on ETM-capable chips  
- **ADC_READ** remains the default and fully supported fallback if digi init fails  

## Limitations

- **ADC1 only**, max **2** pattern channels  
- **ESP32-C3** has no MCPWM — 3PWM uses LEDC; this feature applies to **MCPWM + low-side** builds (e.g. **C6**, S3)  
- Example DRV8302 pins are for classic ESP32 — remap for your board  
- `adc_buffer` layout differs between **ADC_READ** (plain `int`) and **DIGI_*** (`adc_digi_output_data_t` overlay)  

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

**Test Configuration/Setup**:
* Hardware: ESP32C6 + SimpleFoC shield V2.0.1
* IDE: Arduino CLI
* MCU package version arduino-esp32/

### DIFF
--- a/src/current_sense/hardware_specific/esp32/esp32_adc_digi_driver.c
+++ b/src/current_sense/hardware_specific/esp32/esp32_adc_digi_driver.c
@@ -1,0 +1,484 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2021 Felipe Neves
+ *
+ * ADC digi + DMA for SimpleFOC low-side current sense.
+ *
+ * Triggering is selected with esp32_adc_digi_set_trigger():
+ *   - SOFTWARE: always available (ESP32, S2, S3, …).
+ *   - ETM:        ESP32-S3 and newer only; see esp32_adc_etm.c.
+ */
+
+#include "sdkconfig.h"
+#include "esp32_adc_digi_internal.h"
+
+#if SIMPLEFOC_ESP32_ADC_DIGI_SUPPORTED
+
+#include <string.h>
+#include <inttypes.h>
+#include "esp_log.h"
+#include "esp_check.h"
+#include "esp_heap_caps.h"
+#include "esp_clk_tree.h"
+#include "esp_private/regi2c_ctrl.h"
+#include "esp_private/sar_periph_ctrl.h"
+#include "esp_private/adc_share_hw_ctrl.h"
+#include "esp_private/gpio.h"
+#include "soc/adc_periph.h"
+#include "soc/soc_caps.h"
+#include "soc/clk_tree_defs.h"
+#include "hal/adc_hal.h"
+#include "hal/adc_ll.h"
+#include "esp32_adc_digi_driver.h"
+
+#if SOC_CACHE_INTERNAL_MEM_VIA_L1CACHE
+#include "esp_cache.h"
+#endif
+
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
+#define ADC_GET_CHANNEL(p_data)     ((p_data)->type1.channel)
+#define ADC_GET_DATA(p_data)        ((p_data)->type1.data)
+#else
+#define ADC_GET_CHANNEL(p_data)     ((p_data)->type2.channel)
+#define ADC_GET_DATA(p_data)        ((p_data)->type2.data)
+#endif
+
+#define ESP32_ADC_DMA_DESC_ALIGN  4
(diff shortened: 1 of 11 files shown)


uLipe:
@askuric @runger1101001  PTAL this is something that resulted in a HUGE reduction of CPU busy since mostly of the PWM->ADC pipeline is handled by the hardware, PWM generate event to ETM to start the ADC conversion and the DMA stores the channels converted directly in the adc_buffer. with a single interrupt.

Also the ADC Digi is used, allowing faster sample rates and compatibillity across all esp32.

uLipe:
@askuric @runger1101001 CI is happy now!

Happy to hear your comments, I hope with this modification to extract the maximum efficiency from ESP32 ADC infrastructure.

Even though it not imply higher sample rates (> 100K, except on esp32 and esp32c5 which is possible to achieve 2MSPS), it will push the ADC to the maximum supported while

- Improve the precision using the curve fitting
- Makes it to work at full range from 0 to 3.3V
- Does most of the heavy job by HW, MCPWM fires the ETM automagically at each cycle, it triggerts two/three conversions of the ADC and store the results directly on the buffer via DMA.
- And the FoC stack gets only a single interrupt when all this chain is done.

Exceptions apply to esp32 and s2 which uses two interrupts MCPWM fires the scan ADC but the rest of the chain is done via DMA.

AS the PR description says the idea is block the CPU  at minimum for the hard real time things while guarantee the timings and sample point of the current measurement.

uLipe:
@runger1101001 @askuric gentle reminder :)

askuric:
Hi there!
This is awesome!

I am unfortunately not gonna be able to test this anytime soon. Basically not before begining of July. If Richard has some time, he might do it, but I would not count on a quick merge of this one. It is a fondamental change and one that needs to be thoroughly tested. 

I'm excited about this one!

uLipe:
@askuric no hurry, I vote also for better testing, if you think it is worth I can share my bench test harness with my evidence of it is working plus instructions for quick reproductibillity serving as reference.

In any case the current method for sensing currents for esp32 are still present a fallback to avoid the introduction of breaking changes.

PR #546, opened 2026-06-14
stm32: do not declare ADC interrupts unless using SIMPLEFOC_STM32_ADC_INTERRUPT

Moddingear:
This allows the user to declare their own ADC interrupts when they're not in use by SimpleFOC (which they're not by default)


dekutree64:
Looks good. I usually use DMA for the ADC results, but have used interrupts as well, and had to hunt down that function in the library code and disable it. Better to control it with a #define so you don't have to modify the library code.

PR #550, opened 2026-07-06
HardwareTimer liberation

Moddingear:
SimpleFOC relies on some functions from HardwareTimer. 

HardwareTimer register all interrupt for all timers, making timer interrupts unusable on STM32.

HardwareTimer can be disabled by defining HAL_TIM_MODULE_ONLY, but then SimpleFOC is missing some functions.

This makes it so that those function exist when HAL_TIM_MODULE_ONLY is defined.

Should be merged at the same time as https://github.com/simplefoc/Arduino-FOC-drivers/pull/94


Moddingear:
(test doesn't pass because this isn't enabled, if it was, they would pass)

PR #551, opened 2026-07-06
Add printf to SimpleFOCDebug, improve disabled behavior

Moddingear:
Not really used in the library for now, but makes writing stuff easier.


Moddingear:
those where it failed don't know va_start...

PR #552, opened 2026-07-08
fix(teensy): use 12-bit ADC resolution for inline current sense

Armesh26:
## Summary
- Fix Teensy inline current sense ADC conversion constant to match 12-bit ADC reads (4096) instead of 10-bit (1024).

## Context
This matches Teensy `analogReadResolution(12)` behavior and prevents current measurements from being scaled incorrectly.

## Test plan
- [ ] Build example(s) on a Teensy target
- [ ] Verify current-sense readings scale correctly

Made with [Cursor](https://cursor.com)


runger1101001:
Thanks for this. I've changed the branch to dev, as we only accept changes on the dev branch...

looks like a simple change so pending the checks I will merge it.

